Issue #49, opened 2026-08-27
IGDB signals the scan path still cannot see: token-endpoint 429, search-call 401


dgahagan:
Deferred from the v0.22.0 diff review (`claude-F3`, `claude-F4`), both verified against the code and both consciously left out of that release's blast radius. Neither is a regression — each is a gap v0.22.0 narrowed the surrounding area around without closing.

v0.22.0 made a *rejected* Twitch credential distinguishable from a genuine miss. Two adjacent IGDB failures still collapse to `no_match`:

**1. A 429 from the Twitch token endpoint reads as "no match".** `_get_token` returns `None` for any non-auth, non-200 response before any rate-limit predicate runs, so a throttled token exchange is indistinguishable from IGDB having no record. Narrow in practice — Twitch's token endpoint is generous — and the fix threads an `on_rate_limit` callback through `_get_token`'s four consumers, which is a decision table of its own rather than a one-liner. The comment at `items_common.py` in `_scan_upc_game` currently overstates what `on_rate_limit` covers: it is the *search* call that reports a 429, not the token exchange.

**2. A 401 on the IGDB `/games` call reads as "no match".** `search_games` returns `[]` for any non-200 status. A cached token outliving a revoked app is the case that hits it: the exchange succeeded earlier, so `IgdbAuthError` never fires, and the search then fails on a credential that is genuinely dead. The fix needs `_AUTH_STATUSES` applied to the search response *plus* eviction from `_token_cache`, or the next call reuses the same dead token.

Both belong with the cover-picker work — `search_game_art` needs the same treatment, and doing them together avoids threading the same callback through IGDB twice.

Found by the cross-vendor diff review on `feat/issues-42-44-scan-outcome-honesty`; see the archived triage for the full dispositions.

dgahagan:
Half of this shipped in [v0.24.0](https://github.com/dgahagan/shelf/releases/tag/v0.24.0); the other half is deliberately still open, so here is exactly where the line falls.

**Fixed — the token endpoint.** `_get_token` returned `str | None`, so a 429 from Twitch when Shelf renewed its access token had no channel at all and surfaced as "no IGDB match for this barcode". It now returns the same `ProviderResult` record as every other metadata call, and a rate-limited token exchange renders the rate-limit notice.

[...]

Issue #43, opened 2026-08-27
A no-signal (tier 4) scan can be given a confidently wrong film title


dgahagan:
Found during the v0.21.0 test drive (QA Observation 3). Within the letter of that plan's design — hardware was explicitly out of scope — but it produces the worst-looking card in the drive.

## What happens

Scanning `711719541028` (a PlayStation 5 console) gives an honest notice:

> UPC barcode carried no usable title or category signal — filed as DVD / Blu-ray. Change it on the item if that's wrong.

...paired with a specific, confident, wrong title: **"PlayStation: Makers & Gamers - Street Fighter"**. The stored row keeps it.

The log shows why:

```
GET /3/search/movie?query=PlayStation+5+Console  200   (no usable hit)
GET /3/search/movie?query=PlayStation            200   (matched)
```

## Two separable causes

1. **Tier 4's no-signal fallback under Auto is `dvd`** (`app/services/detect.py:203`), so every unrecognised UPC is sent into a film search.
2. **The retail-title ladder descends to a one-word query** (`app/services/upcitemdb.py:103` `search_queries`) even on a scan that has already declared itself low-confidence. "PlayStation" matches a real film on TMDb.

The card therefore says "we found no signal" and names a title anyway.

## Suggested shape

Do not descend the ladder when detection reached tier 4 with **no signal** — keep the scanned product title rather than accept a low-confidence provider title. That leaves an honest notice next to an honest title.

This feeds the deferred "hardware media type" revisit trigger in `plan-issue-36-scan-media-detection.md`; a hardware type would address cause 1, but capping the ladder addresses cause 2 independently and is the cheaper of the two.

dgahagan:
Shipped in [v0.25.1](https://github.com/dgahagan/shelf/releases/tag/v0.25.1).

A scan whose retail title names console hardware is now filed under that title with **no lookup attempted at all**, and the card says so: *"Added with title only — the title names console hardware, so no film or game lookup was attempted."* `711719541028` files as `PlayStation 5 Console` and sends TMDb nothing. `Nintendo Switch Pro Controller` and `Xbox Series X Console 1TB` were the same defect on barcodes the issue did not name, and are fixed by the same change.

[...]

PR #52, opened 2026-08-28
Add optional credentialed Google Books support

martialartistslife:
## Summary

- add optional GOOGLE_BOOKS_API_KEY and encrypted write-only Settings storage
- send configured credentials only through the X-Goog-Api-Key header for ISBN metadata, synopsis, and cover requests
- preserve anonymous Google Books behavior when no key is configured
- add an admin-only, CSRF-protected Settings test action with sanitized responses
- centralize Google Books cover requests in the existing provider service

## Compatibility

- no database schema migration
- environment configuration takes precedence over a stored Settings value, following the existing Shelf secret-setting behavior
- an empty environment value is treated as unset, so a stored key remains effective
- no provider ordering or unrelated metadata fallback behavior changes

## Verification

- 476 focused route/service/settings tests passed
- 175 high-risk credential/settings/module-size tests passed after final structural review
- all 184 Playwright E2E tests passed
- CSRF, Alpine/CSP, service-worker stamp, test badge, test-convention, and whitespace checks passed
- pip-audit found no known vulnerabilities; dependency license inventory completed
- full Windows unit run: 1,920 passed; 11 existing Windows-only failures remain in archive open-file deletion, Unix permission assertions, and worker default text encoding

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,3 +29,7 @@
 # Setting it here means the data directory alone is not enough to decrypt
 # stored credentials. Generate with: openssl rand -hex 32
 # SHELF_ENCRYPTION_KEY=
+
+# Optional Google Books API key. Anonymous Google Books lookups remain enabled
+# when this is unset. The key is sent only in the X-Goog-Api-Key header.
+# GOOGLE_BOOKS_API_KEY=
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 [![Release](https://img.shields.io/github/v/release/dgahagan/shelf)](https://github.com/dgahagan/shelf/releases)
 [![Docker Pulls](https://img.shields.io/docker/pulls/dangahagan/shelf)](https://hub.docker.com/r/dangahagan/shelf)
 [![CI](https://github.com/dgahagan/shelf/actions/workflows/test.yml/badge.svg)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
-[![Unit tests](https://img.shields.io/badge/unit%20tests-1903%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
-[![E2E tests](https://img.shields.io/badge/e2e%20tests-183%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
+[![Unit tests](https://img.shields.io/badge/unit%20tests-1931%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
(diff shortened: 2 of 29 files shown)


dgahagan:
Thanks for this — it's a well-put-together PR, and the security posture in particular is exactly right. Sending the key as `X-Goog-Api-Key` rather than a `?key=` query param matters more here than it might look: our log redaction filter only scrubs query strings, so a header-borne key is out of that blast radius entirely. `test_all_request_variants_send_key_in_header_only` pinning that across all three call sites is the kind of test we want.

I applied the branch to current `main` and ran the full gate. Everything is green:

- `make test` — **1931 passed**
- `make test-e2e` — **203 passed**
- `make check-csrf`, `make check-alpine`, `make check-badges` — pass
- `make css` — no-op, no new Tailwind class

I also checked the things that would have blocked a merge, and none of them bite: no schema/migration change, the new `/api/settings/google-books/test` route is covered by the router-level `require_role("admin")`, the key is Fernet-encrypted at rest and never echoed into the settings HTML, and it can't leak cross-host on a redirect because none of the Google Books calls set `follow_redirects`. Docs are already complete across `.env.example`, `configuration.md`, `integrations.md` and the README tables — thank you for that, it's usually the part that gets left.

**One small thing needed before I can merge:** the two test-count badge lines at the top of `README.md` are the only merge conflict.

```
[![Unit tests](https://img.shields.io/badge/unit%20tests-1931%20passing-brightgreen)]...
[![E2E tests](https://img.shields.io/badge/e2e%20tests-184%20passing-brightgreen)]...
```

Those are generated by `scripts/stamp_test_badges.py` (`make badges`), not hand-maintained, and `main` has moved since you cut the branch — it's at 1903/202 now. Could you revert just those two lines to whatever `main` currently has? I'll run `make badges` after the merge, which stamps the true post-merge numbers (1931/203). Your two Google Books documentation rows further down the file are good as they are — only the badge lines need touching.

Two design notes, both of which are **ours to handle, not asks on you**:

1. `_cover_search_credentials` uses `if not keys:` to mean "this is the book branch". Correct today, since `MEDIA_TYPE_PROVIDERS` only maps `dvd` and `video_game` — just an implicit coupling we'll make explicit later.
2. The new `googlebooks.search_covers` returns `[]` on any non-200, which is the same shape as `igdb.py:178`. We have an open cluster of issues (#49, #43) about exactly that pattern — a provider failure reading as "not found" — so that call site is already on our list to convert. Nothing to change here.

Once the badge lines are reverted I'll merge and it'll ship in the next release. Thanks again.

martialartistslife:
Reverted the two generated README test-count badges to match current upstream main (1903 unit tests and 202 E2E tests), as requested. The Google Books documentation rows remain unchanged.

PR #53, opened 2026-08-28
Fix Hardcover ISBN-10 fallback lookup

martialartistslife:
## Summary

When a Hardcover ISBN-13 lookup misses, the fallback query changes its filter to isbn_10 but currently reuses the original 13-digit value. This converts the ISBN-13 before the retry while preserving the existing behavior for callers that already provide a valid ISBN-10.

The provider order, GraphQL transport, rate-limit callback, and failure behavior are unchanged.

## Tests

- reproduced before the fix: the ISBN-10 query received 9780306406157 instead of 0306406152
- added regression coverage for converted ISBN-13 input and already-valid ISBN-10 input
- 69 ISBN and outbound-provider tests passed
- badge and outbound guard: 31 passed
- CSRF, Alpine/CSP, test-convention, service-worker stamp, module-size, compile, and diff checks passed
- complete Windows unit run exercised 1,933 tests; selected tests passed, while existing Windows archive-file, Unix-permission, and worker-encoding limitations remained

The official badge stamper was run after PR #52 merged, updating the generated counts to 1933 unit and 203 E2E tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 [![Release](https://img.shields.io/github/v/release/dgahagan/shelf)](https://github.com/dgahagan/shelf/releases)
 [![Docker Pulls](https://img.shields.io/docker/pulls/dangahagan/shelf)](https://hub.docker.com/r/dangahagan/shelf)
 [![CI](https://github.com/dgahagan/shelf/actions/workflows/test.yml/badge.svg)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
-[![Unit tests](https://img.shields.io/badge/unit%20tests-1903%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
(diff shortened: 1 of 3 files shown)


dgahagan:
Shipped in [v0.24.0](https://github.com/dgahagan/shelf/releases/tag/v0.24.0). Thank you @martialartistslife — this one was a genuinely good catch.

Worth recording why it survived so long: `lookup_by_isbn` built the ISBN-10 query correctly but passed the unchanged 13-digit value into it, so the retry could never match. Every production caller passes an ISBN-13, which means that branch had matched nothing for the entire life of the file. It stayed invisible because a broken retry and a real miss both came back as the same empty answer — there was no way to tell them apart from the outside.

That is the same defect class as the rest of 0.24.0, which is why the two shipped together. Every metadata client now returns a record saying *what happened* (`found` / `no_match` / `no_credential` / `rejected` / `rate_limited` / `transport_failed`) alongside what it found, so a leg that fails for a fixable reason can no longer be read as "no such book". The regression test for your fix asserts the ISBN-10 value actually sent, not just that a second query was issued — the previous test checked only the sequence of queries, which is precisely why it passed over the bug.

PR #56, opened 2026-08-31
Reject invalid lending notification formats

sudo-rpaisley:
## What this changes

Rejects unknown `notify_format` values submitted to the lending settings endpoint instead of silently coercing them to `ntfy`.

Adds a regression test confirming that an invalid format is rejected before any lending settings are mutated.

## Why

The endpoint currently treats any unrecognised notification format as `ntfy`. A malformed or forged request can therefore change both the notification format and other lending settings while appearing valid.

The settings boundary should reject an unknown enum value rather than silently substitute a different one.

## How it was tested

The equivalent fix and regression test passed the fork's full CI workflow before being rebuilt cleanly on top of upstream `main`.

This PR branch contains only the upstream-targeted change and test; upstream CI can verify it against the current tree.

* [ ] `make test` passes
* [ ] `make test-e2e` passes
* [ ] `make checks` passes

## Notes for review

This is intentionally a small boundary-validation fix.

Normal Settings UI submissions already send one of the supported formats, so there is no template/CSS change and no user-facing workflow change for valid requests.

### DIFF
--- a/app/routers/settings.py
+++ b/app/routers/settings.py
@@ -153,7 +153,9 @@ async def update_lending_settings(
     days = lending_overdue_days.strip() or "28"
     if not days.isdigit():
         return {"ok": False, "message": "Overdue days must be a whole number"}
-    fmt = notify_format if notify_format in FORMATS else "ntfy"
+    fmt = (notify_format or "").strip()
+    if fmt not in FORMATS:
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

The detail that makes this worth fixing: one malformed field also saved the other lending settings in the same request, which the sender never named.

PR #57, opened 2026-08-31
Reject invalid roles when creating users

sudo-rpaisley:
## What this changes

Rejects an unknown role supplied while creating a user instead of silently converting it to `viewer`.

Adds a regression test proving that an invalid role creates no account.

## Why

The user-creation endpoint currently treats any unrecognised role as `viewer`.

That differs from the existing user role-update endpoint, which rejects roles outside `admin`, `editor`, and `viewer` with an `Invalid role` response.

Rejecting malformed input makes the two user-management boundaries consistent and avoids accepting a request while creating an account with different permissions than were requested.

## How it was tested

The equivalent change and regression test passed the fork's full CI workflow.

This branch was rebuilt directly from current upstream `main` and contains only the one-line behaviour change and its focused regression test.

* [x] `make test` passes
* [ ] `make test-e2e` passes
* [ ] `make checks` passes

## Notes for review

No normal UI behaviour changes. Shelf's user-management UI already submits only the supported roles.

This change affects malformed, stale, or manually forged requests only.

### DIFF
--- a/app/routers/auth_routes.py
+++ b/app/routers/auth_routes.py
@@ -150,7 +150,7 @@ async def create_user(
     username = username.strip()
     display_name = display_name.strip() or username
     if role not in ("admin", "editor", "viewer"):
-        role = "viewer"
+        return {"ok": False, "message": "Invalid role"}
     if len(password) < 8:
         return {"ok": False, "message": "Password must be at least 8 characters"}
     if not username or len(username) < 2:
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

Nice consistency argument: the role-*update* endpoint already rejected this input and only create-user coerced it. Two boundaries that disagreed.

PR #58, opened 2026-08-31
Reject invalid share-link mutations

sudo-rpaisley:
## What this changes

Makes share-link mutations reject invalid input instead of silently reporting a different or successful result.

Specifically:

* rejects unknown share scopes instead of silently converting them to `wishlist`;
* returns `404` when attempting to revoke a share link that does not exist;
* adds regression coverage proving invalid share creation does not insert a link and stale revocation does not report success.

## Why

The share-link creation endpoint currently accepts any scope value and silently changes an unknown value to `wishlist`.

That means a malformed, stale, or manually forged request can succeed while creating a different type of share link than was requested.

Similarly, revoking a nonexistent share link currently redirects as though the operation succeeded.

Rejecting these cases makes the mutation endpoints truthful while leaving the normal Settings UI and valid public share-link behaviour unchanged.

## How it was tested

The equivalent change and regression tests passed the fork's full CI workflow.

This branch was rebuilt directly from current upstream `main` and contains only changes to the share-link router and its existing test module.

* [ ] `make test` passes
* [ ] `make test-e2e` passes
* [ ] `make checks` passes

## Notes for review

There is no change to valid share-link creation, public share pages, scope filtering, access control, rate limiting, or revocation of an existing link.

The behaviour change only affects invalid or stale mutation requests.

### DIFF
--- a/app/routers/share.py
+++ b/app/routers/share.py
@@ -7,7 +7,7 @@ field set — never locations, loans, valuations, notes, or ISBNs.
 import secrets
 
 from fastapi import APIRouter, Depends, Form, Request
-from fastapi.responses import HTMLResponse, RedirectResponse
+from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
 
 from app.auth import require_role
 from app.database import get_db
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

Good catch that an unknown share scope became `wishlist` silently: a forged request could create a different kind of link than it asked for. The 404 on revoking a link that never existed is the better half of this.

PR #59, opened 2026-08-31
Validate provider-backed catalogue add inputs

sudo-rpaisley:
## Summary
- reject unknown game platforms instead of silently clearing them
- validate game platforms before any IGDB request is made
- reject blank DVD titles
- reject malformed DVD publish years instead of silently dropping them
- add focused regression coverage for all three boundaries

## Why
The provider-backed catalogue add routes currently handle malformed form values inconsistently. Invalid game platforms can be silently converted to no platform, while invalid DVD years can be silently discarded. Invalid requests should be rejected before mutation or outbound provider work.

## Testing
Adds three focused regressions covering invalid game platforms, blank DVD titles, and malformed DVD publish years.

### DIFF
--- a/app/routers/items_catalog.py
+++ b/app/routers/items_catalog.py
@@ -78,6 +78,17 @@ async def add_game_from_search(
     """Add a video game to the collection from an IGDB search result."""
     templates = request.app.state.templates
 
+    platform = (platform or "").strip()
+    with get_db() as db:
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

Validating the game platform *before* the IGDB request is the part I care about — a bad platform was costing an outbound call as well as storing the wrong thing.

PR #60, opened 2026-08-31
Validate manual catalogue add inputs

sudo-rpaisley:
## What this changes

Makes manual catalogue adds reject malformed input instead of erroring or silently rewriting it.

Specifically:

- rejects malformed publish years before insertion instead of allowing the integer conversion to raise a server error;
- rejects unknown game platforms instead of silently clearing the submitted platform;
- adds regression coverage proving both invalid requests create no catalogue item.

## Why

The manual-add endpoint currently converts `publish_year` with `int(...)` during insertion, so a malformed value can become a server error.

It also silently changes an unknown submitted game platform to no platform at all. A malformed, stale, or manually forged request can therefore succeed while storing different catalogue data than was requested.

Rejecting these cases at the request boundary makes the mutation truthful while leaving normal manual-add behaviour unchanged.

## How it was tested

The equivalent change and regression tests passed the fork's full CI workflow.

This branch was rebuilt directly from current upstream `main` and contains only the manual-add route change and its focused regression test module.

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

There is no change to valid manual catalogue adds, ISBN/UPC handling, duplicate detection, cover handling, locations, or supported game platforms.

The behaviour change only affects malformed publish years and unknown platform values.

### DIFF
--- a/app/routers/items.py
+++ b/app/routers/items.py
@@ -508,8 +508,20 @@ async def manual_add(request: Request, _=Depends(require_role("editor"))):
         upc_code = None
         isbn13 = isbn_svc.to_isbn13(isbn) if isbn else None
         isbn10 = isbn_svc.isbn13_to_isbn10(isbn13) if isbn13 else None
-    pub_year = form.get("publish_year")
-    platform = form.get("platform") or None
+
+    pub_year_raw = (form.get("publish_year") or "").strip()
+    if pub_year_raw:
+        try:
+            pub_year = int(pub_year_raw)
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

`int(publish_year)` at insertion time turning a typo into a 500 is a fair hit. Rejecting the unknown platform rather than quietly clearing it makes this consistent with #59 on the same field.

PR #61, opened 2026-08-31
Reject invalid Hardcover sync intervals

sudo-rpaisley:
## What this changes

Rejects unknown Hardcover sync schedule values instead of silently converting them to `off`.

Specifically:

- accepts the existing supported values `off`, `daily`, and `weekly`;
- returns HTTP 400 for any other interval;
- preserves the currently stored schedule when an invalid value is submitted;
- adds a focused regression proving an invalid request cannot disable an existing schedule.

## Why

The Hardcover schedule endpoint currently treats every unrecognised interval as `off` and then saves it.

That means a malformed, stale, or manually forged request such as `interval=hourly` can silently disable an existing daily or weekly sync schedule while returning the normal success redirect.

Rejecting the invalid value before mutation makes the endpoint truthful and keeps the existing schedule intact.

## How it was tested

The equivalent change and regression test passed the fork's full CI workflow.

This branch was rebuilt directly from current upstream `main` as one commit and contains only the Hardcover schedule route change and its focused regression test.

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

There is no change to normal Hardcover scheduling behaviour. The supported `off`, `daily`, and `weekly` values continue to use the existing redirect and persistence path.

The behaviour change affects unsupported interval values only.

### DIFF
--- a/app/routers/hardcover.py
+++ b/app/routers/hardcover.py
@@ -5,7 +5,7 @@ import re
 
 import httpx
 from fastapi import APIRouter, Depends, Form, Request
-from fastapi.responses import RedirectResponse
+from fastapi.responses import JSONResponse, RedirectResponse
 from starlette.responses import StreamingResponse
 
 from app.auth import require_role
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

Small and exactly right: an unrecognised interval was saved as `off`, so a malformed request could silently disable a schedule someone had set.

PR #62, opened 2026-08-31
Validate Audiobookshelf request boundaries

sudo-rpaisley:
## Summary
- reject malformed Audiobookshelf connection-test URL/token values before saved-credential fallback
- reject malformed excluded-library selections instead of coercing them
- reject unsupported sync intervals instead of silently disabling sync
- add focused regressions proving invalid requests do not mutate settings or make unintended outbound requests

## Testing
Rebuilt as one commit directly from upstream main. Combines two previously full-CI-tested fork fixes that touch the same Audiobookshelf request boundary.

### DIFF
--- a/app/routers/sync.py
+++ b/app/routers/sync.py
@@ -5,7 +5,7 @@ from urllib.parse import urlparse
 
 import httpx
 from fastapi import APIRouter, Depends, Request, Form
-from fastapi.responses import RedirectResponse
+from fastapi.responses import JSONResponse, RedirectResponse
 from starlette.responses import StreamingResponse
 
 from app.auth import require_role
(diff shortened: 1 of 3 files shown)


dgahagan:
Merged, and shipped in v0.27.0 — thank you.

The interval case was the one that mattered to me: an unsupported value used to save as `off`, so a stale form could silently disable a working sync and return the normal redirect. Refusing it and leaving the stored schedule alone is exactly the right shape, and it matches what #61 did for Hardcover. Validating the connection-test URL and token *before* the saved-credential fallback closes a gap I had not noticed: a malformed value was falling through and testing the stored credentials instead.

PR #63, opened 2026-08-31
Validate valuation credential-test requests

sudo-rpaisley:
## Summary
- reject malformed ISBNdb and TMDb credential-test request bodies and non-string keys
- preserve blank/missing key fallback to configured credentials
- ensure invalid supplied values cannot silently test saved credentials
- add focused no-outbound regressions

## Testing


dgahagan:
Merged — thank you.

The one that matters here is that a bad supplied key could fall through and silently test the *saved* credentials instead — that would have been very confusing to debug from the UI.

PR #64, opened 2026-08-31
Validate settings integration-test requests

sudo-rpaisley:
## Summary
- reject non-object notification-test JSON
- reject non-string notification URL/format values
- reject non-string Google Books API-key values
- preserve valid masked-field/configured-credential fallback
- add focused regressions proving malformed requests do not perform outbound work

## Testing
Rebuilt as one commit directly from upstream main from two previously full-CI-tested fork fixes on the same Settings request surface.

### DIFF
--- a/app/routers/settings.py
+++ b/app/routers/settings.py
@@ -77,7 +77,10 @@ async def test_google_books(request: Request):
     if not isinstance(body, dict):
         return {"ok": False, "message": "Invalid request"}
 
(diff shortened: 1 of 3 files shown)


dgahagan:
Merged — thank you.

Rejecting malformed bodies before any outbound work happens is the right ordering, and the masked-field fallback surviving it is the part that needed the tests.

PR #65, opened 2026-08-31
Reject invalid archive import modes

sudo-rpaisley:
## Summary
- reject unknown archive import modes instead of silently coercing them to skip
- apply the same validation to legacy import, preview planning and staged apply
- reject invalid apply modes before staged work is touched
- add focused regression coverage for all three endpoints

## Testing


dgahagan:
Merged — thank you.

Covering legacy import, preview planning and staged apply together was the right call; rejecting an invalid apply mode before any staged work is touched is the part I would have missed.

Issue #54, opened 2026-08-29
ISBN writes can store invalid or contradictory ISBN-10/ISBN-13 pairs


martialartistslife:
## Summary

Shelf can currently persist invalid or internally contradictory ISBN identifiers through supported routes.

Verified against upstream `main` at `5d5934f1c1e4a62e2c84f3b1aedb98ddd6c138ce`.

### 1. Invalid checksum accepted

`app/services/isbn.py::to_isbn13()` normalizes based on shape and length but does not validate the checksum.

- Input: `9780441172710`
- Actual: the value is returned unchanged and can be persisted through supported add/scan paths.
- Expected: the value is rejected because it is not a checksum-valid ISBN-13.

A route-level regression test reproduced persistence of this value.

### 2. Editing an ISBN leaves a stale ISBN-10

Start with an existing item containing:

```text
isbn   = 9780441172719
isbn10 = 0441172717
```

Edit its ISBN field to `054792822X`.

Actual stored pair:

```text
isbn   = 054792822X
isbn10 = 0441172717
```

Expected canonical pair:

```text
isbn   = 9780547928227
isbn10 = 054792822X
```

A route-level regression test reproduced this mismatch.

## Likely cause

- `to_isbn13()` returns any 13-digit value without calling the existing checksum validator (`app/services/isbn.py:17-25`).
- `update_item()` accepts `isbn` as a generic form field and updates `items.isbn` without recomputing or clearing `items.isbn10` (`app/routers/items.py:817-868`).

I searched open and closed upstream issues and pull requests for ISBN validation, checksum validation, normalization, ISBN-10/ISBN-13 synchronization, stale `isbn10`, and item-edit ISBN behavior, and did not find an existing report for this defect. Issue #20 is separate: it concerns UPC values being misfiled as ISBNs and duplicate manual-add behavior.

## Maintainer decision requested

Which fix boundary would you prefer?

**A.** Tighten `to_isbn13()` itself so that it accepts only checksum-valid ISBNs and becomes the canonical validation/conversion helper.

**B.** Preserve `to_isbn13()`'s current permissive semantics, if intentional, and instead enforce checksum validation plus canonical `isbn`/`isbn10` synchronization only at persistence boundaries.

I am happy to prepare a small PR with route-level regression coverage after you confirm the intended approach. No implementation patch is included here.

dgahagan:
[...]

PR #66, opened 2026-08-31
Harden Store queue request and ISBN boundaries

sudo-rpaisley:
## Summary
- reject valid JSON bodies that are not objects
- require checksum-valid ISBNs before metadata lookup or persistence
- canonicalise ISBN-10 input to ISBN-13 while retaining the ISBN-10 pair
- detect duplicates through either canonical ISBN representation
- keep invalid queued scans as invalid instead of creating bare wishlist rows
- correct Store test fixtures to checksum-valid ISBNs

## Testing
Rebuilt as one commit directly from upstream main from two previously full-CI-tested Store boundary fixes.

### DIFF
--- a/app/routers/store.py
+++ b/app/routers/store.py
@@ -96,6 +96,8 @@ async def store_queue(request: Request, _=Depends(require_role("editor"))):
         body = await request.json()
     except Exception:
         return JSONResponse({"error": "Invalid JSON body"}, status_code=400)
+    if not isinstance(body, dict):
+        return JSONResponse({"error": "Invalid JSON body"}, status_code=400)
     isbns = body.get("isbns")
     if not isinstance(isbns, list) or not all(isinstance(x, str) for x in isbns):
         return JSONResponse({"error": "isbns must be a list of strings"}, status_code=400)
@@ -109,15 +111,17 @@ async def store_queue(request: Request, _=Depends(require_role("editor"))):
     results = []
     async with httpx.AsyncClient(timeout=HTTP_TIMEOUT) as client:
         for raw in isbns:
-            isbn13 = isbn_svc.to_isbn13(raw)
-            if not isbn13:
+            pair = isbn_svc.canonical_isbn_pair(raw)
+            if pair is None:
                 results.append({"isbn": raw, "status": "invalid"})
                 continue
+            isbn13, isbn10 = pair
 
             with get_db() as db:
                 existing = db.execute(
-                    "SELECT id, title FROM items WHERE isbn = ? AND media_type = 'book'",
-                    (isbn13,),
+                    "SELECT id, title FROM items WHERE media_type = 'book' AND "
+                    "(isbn = ? OR (? IS NOT NULL AND isbn10 = ?))",
+                    (isbn13, isbn10, isbn10),
                 ).fetchone()
             if existing:
                 results.append({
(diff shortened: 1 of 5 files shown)


dgahagan:
Thanks for this — the validation shape is right, and the fail-closed direction across the whole batch (silent coercion → explicit rejection) is welcome.

One blocker before it can merge: this PR calls `isbn_svc.canonical_isbn_pair(raw)` in `app/routers/store.py`, but `app/services/isbn.py` on `main` has no such function. It exposes `normalize_isbn`, `isbn10_to_isbn13`, `to_isbn13`, `validate_isbn10`, `validate_isbn13` and `isbn13_to_isbn10` — no `canonical_isbn_pair`. No PR in this batch adds it either, so as submitted this raises `AttributeError` on the first ISBN through the queue.

#70 makes the same call in `app/routers/items_csv.py`. I'd guess both were split off a branch where the helper already existed.

Could you add `canonical_isbn_pair` to one of the two PRs, and say which, so the other can be sequenced behind it?

Worth knowing while you shape it: #54 is open on exactly this ground — `to_isbn13()` normalises on shape and length but never checksum-validates, so `9780441172710` persists; and editing an item's `isbn` leaves the old `isbn10` standing, so the stored pair can contradict itself. If your helper already returns a checksum-validated `(isbn13, isbn10)` pair, it may close #54 as well — please say so in the PR if it does, and I'll scope that issue behind it rather than in front.

The other 21 PRs in the batch look independently mergeable and I'll work through them separately.

dgahagan:
Holding this one deliberately, so the red CI is not why it is waiting.

It calls `isbn_svc.canonical_isbn_pair()`, which `main` still does not define — **#79** is the only PR that adds it, and #79 is currently `CONFLICTING` against `main` after the batch merged. If I merged this now, the first ISBN through the route would raise `AttributeError`.

So the order is: #79 rebases and lands, then this goes straight in. Nothing needed from you here unless you would rather move the helper into this PR instead — that would work too, and would unblock both this and #70 without waiting on #79.

Thank you for supplying the helper when I asked, by the way. It closes issue #54 outright.

dgahagan:
Thank you for this one, and sorry it sat — it has been open since 31 August and eight releases have shipped past it, which is squarely my problem and not yours.

I went through the five claims against current `main`. One survives:

| Claim | Status |
|---|---|
| Non-object JSON body accepted | **Real, still there** — `app/routers/store.py:104-107` calls `body.get("isbns")` with no `isinstance(body, dict)` guard, so a bare list or string raises `AttributeError` and surfaces as a 500 instead of the 400 the route intends |
| ISBN not checksum-validated | Already fixed — `store.py:120` calls `canonical_isbn_pair` before any lookup |
| ISBN-10 not canonicalised | Already fixed — `store.py:145`, and `app/services/item_write.py:170-181` writes both columns |
| Dupe detection misses across representations | Not reachable — nothing writes `isbn10` outside the write funnel, so it never exists without its canonical ISBN-13 sibling |

The fifth one I want to explain properly rather than just decline, because you had no way to know and the reasoning is buried in a code comment.

**The bare wishlist row for an unreadable barcode is deliberate.** `store.py:121-144` creates an `"Unreadable barcode — <raw>"` row on purpose and returns `status: "unreadable"`. The reason is at lines 122-129: Store mode is an offline PWA, and the client drops a code from its queue once it has been flushed to the server. If the server answers `"invalid"`, that scan is gone — someone standing in a shop scanned a book and we silently lost it. A junk row they can see and delete is the better failure. Your new test asserts `COUNT(*) == 0` and `status == "invalid"`, which would reverse that.

That is exactly the case the new [CONTRIBUTING.md](https://github.com/dgahagan/shelf/blob/main/CONTRIBUTING.md) now flags — if behaviour looks wrong, ask first, because some of it is deliberate and the reasoning often lives only in a comment. I have used this as the worked example in there, which I hope makes it useful rather than just a decline.

Closing, since the test fixtures here were also superseded by #54. I will land the `isinstance` guard and credit you for the finding in the changelog. The report was good — please keep sending them.

PR #67, opened 2026-08-31
Validate location administration boundaries

sudo-rpaisley:
## Summary
- reject blank and duplicate location names with safe Settings feedback
- report stale update targets instead of database errors
- make deletion of a nonexistent location report failure instead of success
- render only fixed known location error messages
- add focused create/update/delete regression coverage

## Testing
Rebuilt as one commit directly from upstream main from the previously full-CI-tested location CRUD fixes.

### DIFF
--- a/app/routers/locations.py
+++ b/app/routers/locations.py
@@ -1,4 +1,6 @@
-from fastapi import APIRouter, Depends, Request, Form
+import sqlite3
+
+from fastapi import APIRouter, Depends, Form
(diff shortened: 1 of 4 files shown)


dgahagan:
Merged — thank you.

Rendering only fixed known error codes rather than reflecting the query string is the detail I would have missed reviewing this quickly. Deletion of a nonexistent location reporting success was the real bug.

PR #68, opened 2026-08-31
Make tag removal truthful

sudo-rpaisley:
## Summary
- return 404 when the target item no longer exists
- return 404 when the requested tag is not attached to the item
- garbage-collect orphan tags only after an association was actually removed
- preserve unrelated tag associations on stale or forged requests
- add focused regressions

## Testing


dgahagan:
Merged — thank you.

Garbage-collecting orphan tags on the strength of a removal that never happened is a real bug and I had not seen it. Thank you.

PR #69, opened 2026-08-31
Handle scan fetch failures truthfully

sudo-rpaisley:
## Summary
- reject non-2xx responses before consuming recent-scan HTML
- prevent Inventory Missing from rendering HTTP error bodies as successful content
- make camera scans treat HTTP failures as scan failures
- show concise user-facing error feedback
- add an offline regression for all three fetch boundaries

## Testing


dgahagan:
Merged — thank you.

Rendering an HTTP error body as though it were scan content is exactly the class of thing the unit suite cannot see. Three fetch boundaries in one pass.

PR #70, opened 2026-08-31
Harden CSV import boundaries

sudo-rpaisley:
## Summary
- canonicalise imported ISBNs and persist matching ISBN-13/ISBN-10 pairs
- deduplicate ISBN-10 imports against canonical ISBN-13 rows
- reject bad-checksum ISBNs instead of silently importing ISBN-less rows
- reject unsupported import modes and media types before persistence
- correct affected fixtures to checksum-valid ISBNs
- add focused regressions for all boundary cases

## Testing

### DIFF
--- a/app/routers/items_csv.py
+++ b/app/routers/items_csv.py
@@ -15,6 +15,7 @@ from fastapi.responses import JSONResponse
 from starlette.responses import Response
 
 from app.auth import require_role
+from app.config import MEDIA_TYPES
 from app.database import get_db
 from app.routers import items_common
 from app.services import cover_queue
@@ -73,6 +74,13 @@ async def import_csv(request: Request, _=Depends(require_role("admin"))):
 
     form = await request.form()
     mode = form.get("mode", "skip")  # skip or update
+    if mode not in ("skip", "update"):
+        return {
+            "error": "Invalid import mode",
+            "imported": 0,
+            "skipped": 0,
+            "errors": [],
+        }
     to_read_wishlist = form.get("to_read_wishlist") in ("1", "true", "on")
     enrich_covers = form.get("enrich_covers") in ("1", "true", "on")
     csv_file = form.get("file")
@@ -125,8 +133,21 @@ async def import_csv(request: Request, _=Depends(require_role("admin"))):
                     errors.append(f"Row {i}: series_name too long (max {_CSV_MAX_TEXT} chars)")
                     continue
 
-                isbn_val = norm["isbn"]
+                raw_isbn = norm["isbn"]
+                if raw_isbn:
+                    isbn_pair = isbn_svc.canonical_isbn_pair(raw_isbn)
+                    if isbn_pair is None:
+                        errors.append(f"Row {i}: invalid ISBN")
+                        continue
+                else:
+                    isbn_pair = None
+                isbn_val = isbn_pair[0] if isbn_pair else None
+                isbn10_val = isbn_pair[1] if isbn_pair else None
+
                 media = norm["media_type"]
+                if media not in MEDIA_TYPES:
+                    errors.append(f"Row {i}: invalid media_type")
+                    continue
 
                 owned = norm["owned"]
                 if to_read_wishlist and norm["reading_status"] == "want_to_read":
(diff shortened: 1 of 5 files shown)


dgahagan:
Same blocker as #66, flagged there in more detail: this PR calls `isbn_svc.canonical_isbn_pair(raw_isbn)` in `app/routers/items_csv.py`, and `app/services/isbn.py` on `main` has no such function — nor does any other PR in this batch add it. As submitted, any CSV row carrying an ISBN raises `AttributeError`.

Could you add the helper to whichever of #66 / #70 you'd rather carry it, and say which? The other then sequences behind it. Context on the shape — and its overlap with open issue #54 — is in my comment on #66.

The rest of the change here reads well; the `media not in MEDIA_TYPES` guard and the `isbn10` half of the duplicate lookup are both real improvements.

dgahagan:
Holding this one deliberately, so the red CI is not why it is waiting.

It calls `isbn_svc.canonical_isbn_pair()`, which `main` still does not define — **#79** is the only PR that adds it, and #79 is currently `CONFLICTING` against `main` after the batch merged. If I merged this now, the first ISBN through the route would raise `AttributeError`.

So the order is: #79 rebases and lands, then this goes straight in. Nothing needed from you here unless you would rather move the helper into this PR instead — that would work too, and would unblock both this and #66 without waiting on #79.

Thank you for supplying the helper when I asked, by the way. It closes issue #54 outright.

dgahagan:
Thank you — and the same apology as on #66: this has been open since 31 August with eight releases shipping past it. That is a process failure on my side, now addressed (see below).

Checked all four claims against current `main`. **Two are still real**, and they are good catches:

| Claim | Status |
|---|---|
| ISBN-10 not deduped against canonical ISBN-13 rows | **Real** — `app/routers/items_csv.py:128` takes `isbn_val` raw and the dedup query at `:151-154` compares that raw value, so an ISBN-10 row never matches its existing ISBN-13 twin. `UNIQUE(isbn, media_type)` stops the actual duplicate, but the user gets a raw `IntegrityError` string instead of a clean `skipped`, and in `mode=update` the update is skipped entirely |
| Unsupported import modes not rejected | **Real, and the worse of the two** — `items_csv.py:75` defaults `mode` to `skip`, but `:171` only tests `if mode == "skip"`, so any unrecognised value falls through to the **update** branch and overwrites existing rows. Silent, and failing the wrong way |
| ISBNs not canonicalised on import | Already fixed — `insert_item` canonicalises and persists both columns |
| Bad-checksum ISBNs silently imported | Already handled — `item_write.py:178-180` raises `InvalidIsbn`, reported as `Row N: Invalid ISBN` with no row written |

I am closing this rather than asking you to rebase, because the surviving value is about six lines and the rest of the diff is test churn that #54 already landed — a rebase would be almost entirely conflict resolution for two one-line fixes. That is not a good use of your evening.

I will land both fixes and credit you in the changelog. One note if you send something similar again: the mode fall-through is the more interesting bug of the two, and it would have been a great standalone PR on its own — small, one concern, would have merged the same day. The new [CONTRIBUTING.md](https://github.com/dgahagan/shelf/blob/main/CONTRIBUTING.md) explains why that shape wins here, and commits me to giving every open PR a decision at each release instead of letting it rot. Thanks for the report — both of these were worth finding.

dgahagan:
Both fixes shipped today in [v0.33.0](https://github.com/dgahagan/shelf/releases/tag/v0.33.0) — closing the loop on what I promised when I closed this.

The two claims that were still real are both fixed, and both changelog entries credit you and link this PR:

- **The dedup key is now the canonical ISBN-13.** `app/routers/items_csv.py` runs the CSV value through `canonical_isbn_pair()` before the lookup, so an ISBN-10 row, or a hyphenated one, matches the ISBN-13 twin already in the library. It also matches the other direction — an older row that still holds an ISBN-10 is found by a CSV row carrying the 13. No more raw `UNIQUE constraint failed` text reaching the user, and `mode=update` actually runs the refresh it promised.
- **An unrecognised mode is refused whole.** Rejected before the file is read, nothing written.

Two places I went a different way than your diff, both worth saying out loud:

**Matching does not rewrite.** When a CSV row matches an older item whose stored `isbn` is still an ISBN-10, `update` mode refreshes the metadata but leaves the `isbn` column alone. Matching and migrating are different operations, and a silent rewrite of a stored identifier during an import is not something a user asked for. If that column ever gets normalised it should be a deliberate migration, not a side effect of importing an unrelated file.

**The unknown-mode error is HTTP 200 with an `error` key**, not a 4xx. Your PR batch's general direction — replace silent coercion with explicit rejection — is exactly right and is what landed. But this particular route renders into an HTMX card, so a status-code rejection would have been swallowed rather than shown. The body shape is `{"error": "…", "imported": 0, "skipped": 0, "errors": []}`, and the card renders the message. A *missing* `mode` still defaults to `skip`, deliberately — the safe direction.

Verified against a copy of a real 1058-item collection in both modes before release, including your ISBN-10/ISBN-13 case specifically.

Thanks for finding both of these. The mode fall-through in particular was the more dangerous of the two — it overwrote metadata on every matched row — and it would not have been found by anyone reading the happy path.

PR #71, opened 2026-08-31
Harden authentication boundaries

sudo-rpaisley:
## Summary
- use constant password-verification work for unknown usernames without generating a fresh bcrypt hash per request
- serialize first-admin setup and re-check the zero-user invariant inside the write transaction
- handle duplicate-user integrity errors without masking unrelated database failures
- add focused authentication boundary regressions

## Testing
Rebuilt as one clean commit from upstream main from the previously full-CI-tested fork change.

### DIFF
--- a/app/routers/auth_routes.py
+++ b/app/routers/auth_routes.py
@@ -1,6 +1,6 @@
 import logging
+import sqlite3
 
-import bcrypt
 from fastapi import APIRouter, Form, Request, Depends
 from fastapi.responses import RedirectResponse, HTMLResponse
 
@@ -13,6 +13,12 @@ from app.config import get_client_ip
 from app.database import get_db
 
 logger = logging.getLogger(__name__)
+
+# Unknown usernames must do the same one bcrypt verification as known usernames
+# without generating a fresh salt/hash on every request. The hash is created
+# once at process start; the fixed plaintext is deliberately not a valid
+# account credential and only exists to equalise the password-check work factor.
+_DUMMY_PASSWORD_HASH = hash_password("dummy")
 
 router = APIRouter()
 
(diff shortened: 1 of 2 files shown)


dgahagan:
Nineteen of these are merged: #58, #63, #65, #68, #69, #74, #77, #78, #57, #71,
#56, #64, #59, #60, #61, #67, #72, #75, #76. Thank you — this is a lot of careful work, and the theme running
through it is the right one. A write that cannot be honoured should say so
rather than quietly doing something else, and you found that pattern in about a
twenty places I had not looked at.

**One thing you should know, because it will look alarming otherwise: every one
of your PRs shows a red CI run, and none of them is your fault.** The only
failing test is `test_readme_badges_are_current`, which checks that the
test-count badge in README matches what pytest collects. Every PR that adds a
test makes that badge stale, so every PR that adds a test goes red.

There is no way for you to fix it either. If each PR restamped the badge, all of
them would then conflict with each other on that one README line — 21 PRs that
are otherwise disjoint would become 21 that cannot be merged in any order. You
were right not to touch it. I regenerate it once after each batch lands, and
**I have now changed the check so it no longer fails a pull-request build** — it
reports there and passes, while still failing on push to main and locally, where
the person who can run `make badges` actually is. That is live on main, so PRs
you open from here on should not hit it. It cannot retroactively green the
branches already open: re-running them replays the original ref, and I cannot
rebase your branches from this side.

For the record, #76 is the only one of the 25 that passed CI outright — because
it *replaces* a test rather than adding one, so the count never moved. A neat
confirmation of the diagnosis rather than anything you did differently.

**On sequencing.** I am merging in waves rather than all at once, smallest and
most independent first. #79 goes last, because it overlaps seven of the others
and I would rather they land first and shrink it than have them all rebase onto
it. I will comment on #79 separately with a couple of specific asks.

**And thank you for `canonical_isbn_pair()`.** I asked on #66 and #70 what the
missing helper was meant to be; you answered with the implementation, and it
closes issue #54 outright — including the half I had nearly missed, where
editing an item's ISBN left the old `isbn10` standing. Your test asserting
`canonical_isbn_pair("9780441172710") is None` uses the exact value from that
issue's report.

All of it is credited in the changelog under `[Unreleased]`, by name and PR
number, and it will ship that way.

dgahagan:
Merged — thank you.

The per-request bcrypt hash for unknown usernames was mine and it was a poor way to pad timing. Serialising first-admin setup and re-checking the zero-user invariant inside the transaction is a genuine hardening.

PR #72, opened 2026-08-31
Make checkout request boundaries truthful

sudo-rpaisley:
## Summary
- reject due-date values that overflow the supported date range instead of raising a server error
- return 404 when checking in a checkout that does not exist
- return 409 when a checkout has already been returned
- make the check-in update conditional so a loan can only be closed once
- add focused regressions

## Testing
Rebuilt directly from upstream main as an independent checkout-only change.

### DIFF
--- a/app/routers/checkouts.py
+++ b/app/routers/checkouts.py
@@ -1,7 +1,7 @@
 from datetime import date, timedelta
 
 from fastapi import APIRouter, Depends, Form, Request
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you.

The 409 for an already-returned loan is better than what I would have written — I would probably have made it idempotent and lost the information. Making the check-in update conditional so it can only close once is the right shape.

PR #73, opened 2026-08-31
Harden Audiobookshelf sync correctness and diagnostics

sudo-rpaisley:
## Summary
- adopt an existing manually-catalogued same-format ISBN instead of attempting a duplicate insert
- skip duplicate Audiobookshelf same-format ISBN records without aborting the sync
- allow healthy libraries to continue when one Audiobookshelf library times out
- distinguish real updates from unchanged items and avoid rewriting unchanged metadata
- avoid re-downloading covers for unchanged items that already have one
- report useful reasons for skipped items and show the Unchanged count in Settings
- add regressions for ISBN collisions, repeat-sync idempotency and library timeouts

## Testing
This combines three production-tested Audiobookshelf fixes from the fork into one coherent upstream change. The branch is one commit directly from upstream main.

### DIFF
--- a/app/services/audiobookshelf.py
+++ b/app/services/audiobookshelf.py
@@ -24,12 +24,22 @@ def get_excluded_libraries() -> set[str]:
         return set()
 
 
+def _normalise_publish_year(value):
+    """Match SQLite's usual INTEGER coercion so an unchanged year stays unchanged."""
+    if value in (None, ""):
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return value
+
+
 async def sync(abs_url: str, abs_token: str, on_progress=None) -> dict:
     """Sync items from Audiobookshelf. Returns summary stats.
 
     on_progress: optional async callback(current, total, title, status) for progress updates.
     """
-    stats = {"added": 0, "updated": 0, "skipped": 0, "errors": 0}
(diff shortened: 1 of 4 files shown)


dgahagan:
Merged, and shipped in v0.27.0 — thank you. This one got two changelog lines rather than one: the sync-correctness work under **Fixed**, and the **Unchanged** count under **Added**, since that number is the thing someone actually sees.

Three things I want to say specifically. Adopting a hand-catalogued same-format ISBN instead of attempting a duplicate insert is the behaviour I would have wanted from the start — a physical audiobook and its ABS copy are one item, not two. Letting the healthy libraries finish when one times out changes the failure mode from "the sync did nothing" to "this library did nothing", which is a much better question to be asked. And not re-downloading covers for unchanged items removes the one thing that made repeat syncs feel expensive.

The user docs now describe repeat-sync behaviour (`docs/user-guide/integrations.md`) and the troubleshooting page says a *0 added, N unchanged* summary is success rather than a stall. One honest caveat: this went out on the strength of your regressions plus the suite — I have not yet driven it against my own Audiobookshelf server with the new build. That is my next step on the production box, and I will follow up here if anything surprises me.

dgahagan:
Followed up as promised: upgraded my production box to the 0.27.0 image and ran the Audiobookshelf sync against my own server. It completed cleanly, the repeat run reported unchanged items as unchanged, and nothing surprised me. Thanks again for this one.

PR #74, opened 2026-08-31
Add Stats dashboard product-smoke coverage

sudo-rpaisley:
## Summary
Adds a browser-level product journey for the Stats dashboard.

The test seeds a small realistic collection and verifies:
- headline collection KPIs
- all four dashboard charts
- media-type and location breakdowns
- recent-addition content
- drill-down navigation back to the real item
- a clean browser console

This is test-only; there is no application-code change.

### DIFF
--- a/tests/e2e/test_stats.py
+++ b/tests/e2e/test_stats.py
@@ -1,0 +1,118 @@
+"""E2E product-smoke coverage for the Stats dashboard."""
+from datetime import date
+import sqlite3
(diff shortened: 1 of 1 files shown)


dgahagan:
Merged — thank you.

Test-only, and welcome — the Stats dashboard had no browser-level coverage at all. The drill-down back to a real item is the assertion I would have forgotten.

PR #75, opened 2026-08-31
Fix CSP-safe bulk and shortcut interactions

sudo-rpaisley:
## Summary
- make Browse bulk-location Apply work under Alpine's CSP evaluator without weakening the CSP
- replace inert inline shortcut-modal handlers with external script listeners
- allow Escape to close the shortcut-help modal
- add real browser coverage for bulk move and shortcut controls
- assert the shortcut journey produces no CSP violations

## Testing
Combines the two related CSP interaction fixes from the fork. Exact upstream diff: app.js +37, browse.js +8/-1, one browser test file and the CSP smoke extension.

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -107,6 +107,14 @@ document.body.addEventListener('showToast', function(e) {
 
 // --- Keyboard shortcuts ---
 document.addEventListener('keydown', function(e) {
+    if (e.key === 'Escape') {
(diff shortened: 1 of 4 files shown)


dgahagan:
Merged — thank you.

Fixing bulk Apply under the CSP evaluator *without* weakening the CSP is the constraint that matters here, and asserting the journey produces no CSP violations is a better test than checking the button works.

PR #76, opened 2026-08-31
Reject stale Photo Intake locations

sudo-rpaisley:
## Summary
- introduce a small shared write-target validator for location IDs
- preserve the existing zero/negative no-location sentinel
- reject a stale positive Photo Intake location before integration settings, HTTP clients, metadata lookups or inserts
- return a stable user-facing error rather than leaking a SQLite foreign-key failure
- add regressions for both weak and ISBN-first Intake paths

## Testing
Exact upstream diff: Intake +14/-6, new 30-line helper, and the existing Intake regressions updated to prove no provider work occurs.

### DIFF
--- a/app/routers/intake.py
+++ b/app/routers/intake.py
@@ -18,12 +18,14 @@ from app.services import authors as authors_svc
 from app.services import national
 from app.services.title_match import titles_agree
 from app.services.item_write import insert_item
+from app.services.write_targets import UnknownLocationError, validated_location_id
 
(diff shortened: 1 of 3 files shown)


dgahagan:
Merged — thank you.

The stable user-facing error instead of a leaked SQLite foreign-key failure is the user-visible half, and rejecting before any HTTP client or metadata lookup runs is the expensive half. `write_targets.py` also arrives identically in #79 — same bytes — so the two merge together without conflict.

PR #77, opened 2026-08-31
Align Viewer UI with backend permissions

sudo-rpaisley:
## Summary
- hide item-detail lending/check-in and Hardcover-push mutations from Viewer accounts while preserving loan context
- hide Editor-only Series mutations from Viewers while keeping read-only series and completeness features
- keep Discover searchable for Viewers without showing wishlist mutations or an Admin-only Settings link
- preserve Viewer-permitted reading-status and read-only behaviour
- add a real lend-to-return browser journey and a Viewer-account journey covering item, Series and Discover
- add focused Viewer/Editor Discover rendering regressions

## Why
Several controls were rendered for Viewer accounts even though their endpoints require Editor or Admin. The UI should mirror the existing backend permission contract.

### DIFF
--- a/app/templates/discover.html
+++ b/app/templates/discover.html
@@ -6,8 +6,12 @@
 
     {% if not has_hardcover %}
     <div class="bg-shelf-card rounded-xl border border-shelf-border p-8 text-center">
+        {% if user and user.role == 'admin' %}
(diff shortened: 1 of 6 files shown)


dgahagan:
Merged — thank you.

This one unblocks a roadmap item I had parked: the item detail page offers book-shaped controls on every media type, and I was waiting on your changes to that template before scoping it. I will scope it against the merged file.

PR #78, opened 2026-08-31
Make valuation report printing CSP-safe

sudo-rpaisley:
## Summary

* replace the valuation report's inline `onclick="window.print()"` handler with a CSP-safe external script
* keep the visible **Print / Save as PDF** control working under Shelf's existing `script-src 'self'` policy
* add a real browser product-smoke journey that verifies valuation contents, manual-value handling, and the print action itself

## Why

The report currently renders correctly but its main output action is silently blocked by the application's Content Security Policy because inline script handlers are not permitted.

This was found while continuing the product-smoke audit: unit coverage proves the report HTML and totals, but only a browser-level click exposes that the visible print button is inert.

## Testing

This branch is one commit rebuilt directly from current upstream `main` (`a3910c1`) and contains only the valuation-report fix plus its focused E2E regression.

### DIFF
--- a/app/templates/valuation_report.html
+++ b/app/templates/valuation_report.html
@@ -18,12 +18,13 @@
         a { text-decoration: none; color: inherit; }
     }
 </style>
+<script src="/static/js/valuation-report.js" defer></script>
 {% endblock %}
 {% block content %}
 <div class="max-w-4xl mx-auto">
     <div class="flex items-center justify-between mb-2">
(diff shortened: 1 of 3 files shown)


dgahagan:
Merged — thank you.

A visible button that CSP silently makes inert is the perfect example of why product-smoke coverage earns its place. Unit tests proved the report's HTML and totals the whole time.

Issue #20, opened 2026-08-19
Manually adding the same unresolvable barcode twice returns a 500


dgahagan:
Scanning a barcode that no metadata provider resolves, adding it manually, then scanning that same barcode again offers the manual-add form a second time — and submitting it returns **HTTP 500** instead of a "duplicate" result.

Pre-existing (not new in 0.5.0); found while test-driving 0.5.0. Only the **UPC path** (DVD/Blu-ray, video games) is affected — see below for why ISBNs are safe.

### Steps to reproduce

1. Scan tab → Media Type **DVD / Blu-ray**.
2. Enter a barcode nothing will resolve, e.g. `888888888866` → "Not found — add manually below".
3. Give it a title, **Add to Collection**. Item is created (200).
4. Scan the **same** barcode again → "Not found — add manually below" appears again, rather than "duplicate".
5. Submit the manual form again → **500**.

Verified end to end against 0.5.0; the two `POST /api/items/manual` calls return `200` then `500`.

```
sqlite3.IntegrityError: UNIQUE constraint failed: items.isbn, items.media_type
  File "/app/app/routers/items.py", line 537, in manual_add
```

### Cause

Two separate defects that line up:

1. **The duplicate check reads a different column than the manual add writes.**
   `_scan_upc` looks for an existing item with `WHERE upc = ? AND media_type = ?` (`app/routers/items.py:1651`), but `manual_add` stores the scanned code in **`items.isbn`** — passed through `isbn_svc.to_isbn13()`, which zero-pads a 12-digit UPC-A to 13 — and never populates `items.upc`.

   So the row created at step 3 is `isbn='0888888888866'`, `upc=NULL`, and the check at step 4 cannot see it.

2. **`manual_add` doesn't handle the constraint.** The second insert violates `UNIQUE(isbn, media_type)` (`app/database.py:34`) and the `sqlite3.IntegrityError` propagates uncaught, so the user gets a 500 page instead of a result card.

ISBN scans are unaffected: that path dedupes via `_find_item_by_barcode()` (`items.py:371`), which normalises with the same `to_isbn13()` that `manual_add` stores, so the lookup matches.

### Expected

Step 4 should report the existing item as a duplicate, exactly as scanning a known barcode does. Failing that, step 5 should never return a 500.

### Suggested fix

Either half stops the 500; doing both is better:

[...]

PR #79, opened 2026-08-31
Roll up remaining audit integrity fixes

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test
## The problem

The test is

different correct responses:

## Proposed split

objection does not apply.

than no test.

## Residual risk

## Done when

      stated retirement trigger.

sudo-rpaisley:
Complete the remaining unsent product-smoke and mutation-boundary audit work in one reviewable commit.

Catalogue and lending integrity
- validate ISBN checksums and canonical ISBN pairs across catalogue writes
- harden bulk updates, merge operations, scan targets, borrower mutations and checkout targets
- preserve related catalogue data during merges and reject unsafe merge states
- serialize checkout creation so concurrent requests cannot create contradictory active loans

Browse and write targets
- make Browse selection and bulk controls role-aware
- report partial and failed bulk deletes truthfully instead of always showing success
- reject stale catalogue locations before provider lookups or inserts
- reject cover selection for missing items before network or filesystem work

Item editing
- validate numeric, enum, ownership, location and title fields before mutation
- reject catalogue conflicts cleanly and avoid partial cover writes
- reject invalid quick reading-status values instead of silently clearing state

Integration request boundaries
- reject malformed IGDB credential-test bodies before outbound requests
- validate Hardcover credential-test and add-to-shelf request shapes before network or catalogue work
- retain the Hardcover invalid-schedule guard needed by the shared router state

Platform administration
- report blank names, slug collisions and stale platform deletions truthfully
- render fixed whitelisted Settings feedback for platform and borrower mutation errors

Testing
- add focused unit regressions for catalogue, merge, lending, write-target, item-edit, platform, IGDB and Hardcover boundaries
- add browser regressions for role-aware Browse controls and failed bulk-delete feedback

This commit is rebuilt directly on upstream 0.26.0 (a3910c1). It intentionally excludes the unrelated changes already proposed in upstream PRs #56-#78. Two small shared-file dependency pieces overlap existing reviews: Browse's CSP-safe parseInt helper from #75 and the Hardcover invalid-schedule guard from #61.

## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.

### DIFF
--- a/app/routers/checkouts.py
+++ b/app/routers/checkouts.py
@@ -1,7 +1,7 @@
 from datetime import date, timedelta
 
 from fastapi import APIRouter, Depends, Form, Request
-from fastapi.responses import RedirectResponse
+from fastapi.responses import JSONResponse, RedirectResponse
 
 from app.auth import require_role
 from app.database import get_db, get_setting
@@ -47,12 +47,26 @@ def get_overdue_loans(db) -> list[dict]:
     return [dict(row) for row in rows]
 
 
+def _borrower_settings_error(code: str) -> RedirectResponse:
+    return RedirectResponse(url=f"/settings?borrower_error={code}", status_code=303)
+
+
 # --- Borrowers ---
 
 @router.post("/borrowers")
 async def create_borrower(name: str = Form(...), _=Depends(require_role("admin"))):
+    clean_name = name.strip()
+    if not clean_name:
+        return JSONResponse(
+            {"ok": False, "message": "Borrower name is required"}, status_code=400
+        )
     with get_db() as db:
-        db.execute("INSERT OR IGNORE INTO borrowers (name) VALUES (?)", (name.strip(),))
+        existing = db.execute(
+            "SELECT id FROM borrowers WHERE name = ?", (clean_name,)
+        ).fetchone()
+        if existing:
+            return _borrower_settings_error("duplicate")
+        db.execute("INSERT INTO borrowers (name) VALUES (?)", (clean_name,))
     return RedirectResponse(url="/settings", status_code=303)
 
 
@@ -66,14 +80,14 @@ async def delete_borrower(borrower_id: int, _=Depends(require_role("admin"))):
     matching what deleting a location or a platform already does.
     """
     with get_db() as db:
-        # Take the write lock *before* reading the guard. sqlite3 opens no
-        # transaction for a bare SELECT, so without this the active-loan
-        # count is read outside any lock and a checkout committed between
-        # that read and the DELETE below would be destroyed as "history".
-        # The foreign key used to make that interleaving fail safe; the
-        # cascade removes that accidental protection, so the lock replaces
-        # it. Read, decide, and write are now one serialized unit.
+        # Take the write lock *before* reading either existence or the active-
+        # loan guard. This serializes the decision with checkout creation.
         db.execute("BEGIN IMMEDIATE")
+        borrower = db.execute(
+            "SELECT id FROM borrowers WHERE id = ?", (borrower_id,)
+        ).fetchone()
+        if not borrower:
+            return _borrower_settings_error("missing")
         active = db.execute(
             "SELECT COUNT(*) as c FROM checkouts WHERE borrower_id = ? AND checked_in IS NULL",
             (borrower_id,),
@@ -97,11 +111,27 @@ async def checkout_item(
     _=Depends(require_role("editor")),
 ):
     """Check out an item to a borrower."""
-    templates = request.app.state.templates
     due = (date.today() + timedelta(days=due_days)).isoformat() if due_days > 0 else None
 
     with get_db() as db:
-        # Check not already checked out
+        # Guard and insert must be one serialized write decision. Without a
+        # write lock, two requests can both observe no active checkout before
+        # either INSERT commits and create contradictory simultaneous loans.
+        db.execute("BEGIN IMMEDIATE")
+        item = db.execute("SELECT id FROM items WHERE id = ?", (item_id,)).fetchone()
+        if not item:
+            return JSONResponse(
+                {"ok": False, "message": "Item not found"}, status_code=404
+            )
+
+        borrower = db.execute(
+            "SELECT id FROM borrowers WHERE id = ?", (borrower_id,)
+        ).fetchone()
+        if not borrower:
+            return JSONResponse(
+                {"ok": False, "message": "Borrower not found"}, status_code=404
+            )
+
         active = db.execute(
             "SELECT id FROM checkouts WHERE item_id = ? AND checked_in IS NULL", (item_id,)
         ).fetchone()
--- a/app/routers/hardcover.py
+++ b/app/routers/hardcover.py
@@ -5,7 +5,7 @@ import re
 
 import httpx
 from fastapi import APIRouter, Depends, Form, Request
-from fastapi.responses import RedirectResponse
+from fastapi.responses import JSONResponse, RedirectResponse
 from starlette.responses import StreamingResponse
 
 from app.auth import require_role
@@ -31,8 +31,17 @@ HC_STATUSES = {
 @router.post("/test")
 async def test_hardcover(request: Request, _=Depends(require_role("admin"))):
     """Test a Hardcover API token."""
-    data = await request.json()
-    token = data.get("token", "").strip()
+    try:
+        data = await request.json()
+    except Exception:
+        return {"ok": False, "message": "Invalid request body"}
+    if not isinstance(data, dict):
+        return {"ok": False, "message": "Invalid request body"}
+
+    raw_token = data.get("token")
+    if raw_token is not None and not isinstance(raw_token, str):
+        return {"ok": False, "message": "Invalid request body"}
+    token = (raw_token or "").strip()
     if not token:
         # Masked field posts empty — test the stored token instead
         with get_db() as db:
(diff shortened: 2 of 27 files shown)


dgahagan:
Thanks for this one — and for #80, and for landing `canonical_isbn_pair()`, which
answers the question I asked on #66 and #70. That helper closes issue #54 outright:
the checksum validation is the half I filed it for, and rewriting `isbn10` alongside
`isbn` on edit is the half I nearly missed. `tests/test_isbn.py` asserting
`canonical_isbn_pair("9780441172710") is None` is the reporter's exact repro value.

Three things before I merge this one, and one note on sequencing.

**1. It removes 86 comment lines from `app/routers/items.py` and adds none.**
Net across non-test source is −74. Some of what goes is load-bearing — this block
in `_find_duplicate_item()`, for instance:

```python
# Rows written before #20 kept the barcode in items.isbn, zero-padded
# by to_isbn13() — the same canonical form normalize_upc() produces,
# so this matches them exactly. Migration 21 re-files them, but it
# skips any row whose move would collide, and an older instance may
# still share the database. No real ISBN can land here: ISBN-13 is
# always 978/979, which detect_barcode_type() classifies as an ISBN.
```

That paragraph is the reason the fallback lookup below it is safe, and it is not
recoverable from the code. Same for the "Guard 3 of 3" comment in `manual_add()`
explaining why `is_valid_media_type` is called at that specific boundary. Could
you restore the comments in `items.py`? If any of them are now wrong because the
code moved under them, correcting them is very welcome — it is only the silent
deletion I want to avoid.

**2. Trailing blank lines are stripped at EOF** (`items.py`, `checkouts.py`).
Same thing happened in #71 on `auth_routes.py`. Harmless, but it puts unrelated
noise in the diff — could you drop those hunks?

**3. This one is not in the batch security review.** I reviewed #56–#78 together
on 2026-08-31; #79 and #80 arrived after and are not covered. That is on me to
redo, not work for you — flagging it so you know why this one sits longer than
the others did.

**On sequencing.** I am merging the smaller PRs first, so this one will need a
rebase. Specifically:

- **#61** is fully contained here — both `hardcover.py` hunks are verbatim and
  the new test file is byte-identical. I am merging #61 on its own rather than
  closing it; your hunks here will then land as no-ops.
- **#76**'s `app/services/write_targets.py` is byte-identical to yours, so that
  add/add resolves itself.
- **#75**'s `browse.js` hunks are contained here too. One thing to keep when you
  rebase: **your `tests/e2e/test_bulk_actions.py` is the better version** and
  #75's is what is on `main` now. Arming `expect_navigation()` before the click,
  instead of waiting for `networkidle` after it, is a real flake fix — please
  keep yours over the merged one. That exact race bit us elsewhere in the
  meantime: `test_csv_import` failed intermittently on #76's CI, a branch that
  touches nothing on the CSV path. I have fixed that one and filed the wider
  pattern; your diagnosis is what made it recognisable.
- **#60** is the one real collision: it rewrites the `pub_year` / `platform` lines
  that sit three lines below your ISBN rewrite in `manual_add()`. Expect a
  conflict there and nowhere else in that file.

So: #59, #60, #61, #67, #72, #75, #76 went in first, and this now needs a rebase
onto them. GitHub reports the conflict in `app/routers/checkouts.py`, against #72.

I cannot do it from this side — "allow edits by maintainers" is off on this PR, so
the rebase has to come from you. **Two other PRs are waiting behind it**: #66 and
#70 both call `canonical_isbn_pair()`, which only this PR adds, so they cannot
merge until this does.

If you would rather split this up, that would also work and might be easier: the
`canonical_isbn_pair()` half on its own would unblock #66 and #70 immediately,
and the rest could follow at whatever pace suits you.

One piece of good news for the rebase: **the badge check no longer fails
pull-request builds.** It reports and passes there now, and still enforces on
push to main. So once you rebase onto current `main`, this should get a genuinely
clean CI run rather than the red every one of your PRs has shown so far.
#80 needs the same treatment — it conflicts on the README badge line, which is a
generated file, so just take whatever `main` has for it.

dgahagan:
@sudo-rpaisley — an update on this one, because part of it has shipped and the rest now needs a decision from you rather than from me.

**What landed, verbatim.** `canonical_isbn_pair()` and its seven tests (plus the EAN-13 passthrough pin) are in [v0.28.0](https://github.com/dgahagan/shelf/releases/tag/v0.28.0), taken from this PR unchanged and committed under your authorship — `0c77f3b` in the internal history, and part of the squashed `97899ae` here. `to_isbn13()` kept its position and gained your docstring saying it is deliberately permissive. That helper is now the strict rule at every user-facing ISBN write boundary in the app, so it did rather more work than its size suggests. Thank you — it was the right shape, and it saved me writing a worse one.

**Where 0.28.0 took a different route.** This PR validates at the call sites; the release moved the check *into* the write layer instead. `app/services/item_write.py` now holds a single `validate_item_fields()` that both `insert_item()` and two new update funnels run, so add, edit, bulk edit, merge, CSV import, the store queue, photo intake, the syncs and archive import get the same rules and the same error type without each route repeating them. A test (`test_only_item_write_updates_user_fields`) now fails if a raw `UPDATE items SET` on a user-supplied field appears outside that module, which is what keeps the funnel from being routed around later.

The practical consequence for this PR: the areas it hardens — item-edit field validation, merge safety, stale write targets, scan targets — are covered by that funnel now, reached from a different direction. `write_targets.py` was extended rather than replaced, so your stale-location idea is in there.

**What is still genuinely open here.** This PR is rebuilt on 0.26.0 and the tree has moved twice since, so I do not want to guess which parts you still consider unshipped. Reading it against current `main`, these look like they have no equivalent yet:

- serialising checkout creation so concurrent requests cannot create two contradictory active loans;
- truthful reporting of partial and failed bulk deletes;
- the platform-administration boundaries (blank names, slug collisions, stale deletions) and the whitelisted Settings feedback for them;
- the IGDB and Hardcover credential-test request-shape validation, if #61–#64 did not already cover it.

If you agree, splitting those out onto current `main` as a smaller PR would be much easier for both of us to review than rebasing this one — and the checkout serialisation in particular is the piece I would take first, since it is a correctness bug and not a validation gap. If you would rather rebase the whole thing, that is fine too; say so and I will hold the branch.

One thing I owe you: the release notes and the changelog credit you by name for `canonical_isbn_pair`, but the wider design of the funnel is not attributed to this PR, and I want to be clear that the *idea* of pushing these checks to a single enforced boundary came out of reading it.

dgahagan:
First: thank you. You have sent 21 merged PRs into this project and this one contains two of the most valuable findings anyone has reported — I have filed both so they do not get lost, credited to you: **#86** (merging items silently destroys the merged row's loans, tags and links, via three `ON DELETE CASCADE` declarations) and **#87** (a legacy row with a bad-checksum ISBN cannot be saved at all, so the user is simply stuck). #86 is silent data loss and I would not have found it without you.

I am closing this PR, and I want to be straight about why, because a good part of it is my fault rather than yours.

**What happened.** About 60% of the source changes here have already landed by other routes since you branched — the ISBN checksum and canonical-pair work is now centralised in `app/services/item_write.py`, and `write_targets.py` grew the full `ItemValueError` hierarchy. You could not have known that: I do most of the work in private branches and this repo only sees squashed release commits, so there was no way to see what was in flight. That is a problem with how I publish, not with how you work, and I have started fixing it — see the new [CONTRIBUTING.md](https://github.com/dgahagan/shelf/blob/main/CONTRIBUTING.md), which now explains the cadence and commits me to giving every open PR a decision at each release instead of letting it sit.

**Three specific things that stop this landing as one commit:**

1. It deletes 86 comment lines from `app/routers/items.py` and adds none — the #20 re-file rationale in `_find_duplicate_item`, the "Guard 3 of 3" note in `manual_add`, the G47 cascade comments in `scan_isbn`. Those paragraphs are not recoverable from the code.
2. `UnknownLocationError` subclasses bare `ValueError` here, which breaks every `except ItemValueError` in the current tree.
3. `create_borrower` and `platforms.create_platform` are rewritten from `INSERT OR IGNORE` into read-then-write **without** `BEGIN IMMEDIATE`, against `UNIQUE` columns. That converts a race-free insert into the exact bug class described in `GOTCHAS.md` G18 — a lost race becomes an uncaught `IntegrityError` and a 500.

Your `checkout_item` change is the opposite: it uses `BEGIN IMMEDIATE` correctly, and it is a **genuine new G18 instance that my own in-flight work does not cover** — that plan explicitly scopes `checkouts.py` out. There is no unique constraint behind active loans, so two concurrent POSTs really can create two contradictory ones. I want that fix.

**If you would like to keep going, here is the split, and you get first refusal on any of it:**

| Piece | Scale | Ready? |
|---|---|---|
| 1. Legacy-ISBN edit escape hatch (#87) | ~14 lines | Now |
| 2. Role-aware Browse controls + honest bulk-delete reporting | ~50 lines | Now |
| 3. Route-existence 404s (`delete_item`, `update_item`, cover-write ordering) | ~30 lines | Now |
| 4. Merge integrity (#86) — write fresh, do not port the hunk | ~130 lines | Now |
| 5. `checkout_item` serialization | ~40 lines | Now |
| 6. hardcover/igdb request validation | ~50 lines | After #83 lands — it rewrites `add_hardcover_to_shelf` |

Small and separate is genuinely faster here, and pieces 1 and 5 are close to ready as they stand. Tell me which you want and I will leave them alone; anything you do not want, I will pick up and credit you in the changelog either way. Sorry this one cost you more than it should have.

dgahagan:
@sudo-rpaisley — the bug you found here is fixed as of [v0.42.2](https://github.com/dgahagan/shelf/releases/tag/v0.42.2), and you should know what happened to it, since this PR was closed back in September.

Your diagnosis was right and it is what the fix is built on: the edit form re-posts every control on every save, so a stored ISBN that predates the check-digit validation bounces the save no matter what the user actually changed. That framing is what made it a tractable issue (#87) instead of a vague report. I measured it afterwards — four rows in my own collection, all of them ASINs an older Audiobookshelf sync had put in the ISBN field.

**I took a different route than your patch, and it is worth saying why.** Your `static/js/item_edit.js` hunk disabled the `#isbn` control at submit when `value === defaultValue`, so the browser omitted it. That works in a browser, but it fixes the client and leaves the server contract broken: the unit harness that models a rendered form (`_rendered_form`) and any non-JS client still bounce, so the bug could only be pinned by an E2E test rather than by the suite that runs on every change. It also leaves a disabled control behind — a bfcache restore after a save (browser Back) brings the ISBN input back greyed out and uneditable.

What shipped instead does it server-side: the route compares the submitted identifier against the stored one and drops the field from the write when it is unchanged *and* invalid, letting the existing "a field an update does not carry is not validated" rule handle the rest. Same outcome, pinnable by the unit suite, no client change at all.

Two things I added on top of your version: the form now shows a note under the field so you can see the stored value is not valid rather than it silently passing through, and `upc` gets the identical treatment, since it had the same trap armed one step earlier in the route.

One piece of your PR I did *not* take and have not done: the 404 on a missing item in `update_item`. That is a separate response-shape change and needs its own test; the row read this fix adds makes it close to a one-liner if you still want it.

Thanks for the report and for the detail in the thread — it saved real work.

PR #80, opened 2026-09-01
Add Audiobookshelf browser URL

sudo-rpaisley:
## What this changes

Add an optional browser-facing Audiobookshelf URL while keeping the existing Audiobookshelf URL as Shelf's internal/API endpoint.

- add an `abs_public_url` setting for user-facing Audiobookshelf links
- keep sync, library discovery, cleanup and other API traffic on `abs_url`
- use the browser URL for direct and linked-item Listen/Read links
- fall back to `abs_url` when the browser URL is blank
- trim whitespace/trailing slashes and reject invalid non-HTTP(S) URLs
- add focused regression coverage for persistence, validation, fallback and rendered links
- refresh the generated test-count badge, Tailwind CSS and service-worker version

## Why

Shelf may run in Docker or on a LAN where it can reach Audiobookshelf through an internal address such as `http://audiobookshelf:80`, while a user's browser reaches the same server through a reverse proxy such as `https://audiobooks.example.com`.

Using one URL for both purposes makes the generated Listen/Read links unusable in those deployments. A separate optional browser URL keeps internal API traffic private while producing links the user's browser can actually open.

Leaving the new setting blank preserves the existing behaviour.

## How it was tested

The branch is a single commit directly on current upstream `main` (`a3910c1`) and was validated using Shelf's own upstream-era CI workflow:

- [x] `make test` passes
- [x] `make checks-fast` passes
- [x] `make test-e2e` passes
- [x] `make css` was regenerated and the committed CSS/service-worker output passes the reproducibility check

The focused tests cover saving and reloading the browser URL, preserving the API URL, rejecting invalid values, clearing the override, fallback behaviour, direct ABS links and linked physical/digital copies.

## Notes for review

This feature was split out of the larger audit roll-up PR #79 so the two reviews no longer overlap. PR #79 has been reset to its original non-Audiobookshelf-browser-URL diff.


dgahagan:
Merged 21 of the batch today; this one and #79 are the two left that I cannot merge from here.

**#80 is `CONFLICTING`** — on `README.md`, and only on the test-count badge line. That is a generated file (`make badges`), and it went stale because the 21 merged PRs added tests. So the resolution is just to take whatever `main` has for that line; there is no real conflict in your change.

Two things that should make the rebase easy:

- **The badge check no longer fails pull-request builds.** It reports and passes there now, and still enforces on push to main. Every PR of yours so far has shown a red CI run for that one reason and nothing else — that stops once you are rebased onto current `main`.
- `app/templates/settings.html` moved under you (#67 and #79's area), so expect to re-apply your `abs_public_url` card against the current file.

No hurry. #79 is the one with people waiting behind it — #66 and #70 both need its `canonical_isbn_pair()`.

dgahagan:
Thank you — this one I want, and it is the only open PR I am asking to rebase rather than closing.

The problem is real and nothing on `main` addresses it. `get_playback_url` (`app/services/audiobookshelf.py:318-320`) builds a user-clickable `href` from the same `abs_url` used for server-side API calls, and it renders straight into `app/templates/item_detail.html:134`. Anyone whose Shelf reaches Audiobookshelf over a Docker network or an internal hostname gets a "Listen" link that cannot work from their browser. There is no `abs_public_url` concept anywhere in the tree.

I also want to correct something before you see it elsewhere: I initially suspected you had hand-edited the generated assets, and you had not. `static/css/app.css` differs by exactly the one `.mt-8` class your new fragment introduces, and `static/sw.js` is the matching `SW_VERSION` restamp — you ran `make css` correctly. My apologies.

**What a rebase needs:**

1. Rebase onto current `main`. Only three of your eight files conflict: `README.md`, `app/services/audiobookshelf.py` (`get_playback_url` moved to :318) and `static/sw.js`. The other five apply cleanly.
2. **Drop `README.md`, `static/css/app.css` and `static/sw.js` from the diff and regenerate** — `make css` and `make badges`. Not because you edited them wrongly, but because the service-worker stamp is a digest over the whole precache set: replaying your old stamp onto a moved `main` produces one that fails `make check-sw-version`.
3. Read `abs_public_url` once at the call site in `app/routers/pages.py:215`, beside the existing `get_setting(db, "abs_url")`, and pass it into `get_playback_url`. As written it opens its own `with get_db()` per call, and one caller is a list comprehension over linked items (`pages.py:222`) — that is an N+1 of nested connections inside an in-flight transaction.
4. Consider adding `"abs_public_url"` to `_INTEGRATION_KEYS` in `app/routers/settings.py:19` plus one input in the existing ABS form, instead of the new route and separate settings card. Same result, and it puts the field next to the URL it modifies. Your call — if you prefer the dedicated card, say why and I will take it as-is.

Also needs a `CHANGELOG.md` entry under `[Unreleased]` and a line in `docs/configuration.md`.

Your fragment is clean on the CSP and CSRF rules, and the `x-show` pattern matches the existing tabs — nothing to change there.

Sorry this sat since 1 September. That is now addressed: see the new [CONTRIBUTING.md](https://github.com/dgahagan/shelf/blob/main/CONTRIBUTING.md), which explains the release cadence and commits me to giving every open PR a decision at each release. If you would rather not do the rebase, say so and I will finish it and keep you as the author.

dgahagan:
Heads-up: [v0.32.0](https://github.com/dgahagan/shelf/releases/tag/v0.32.0) just went out, so `main` has moved under you again. Rebase onto that tag rather than onto whatever your branch remembers.

Nothing has changed about the decision from my last comment — this is still the one open PR I want and am asking to rebase rather than close. The problem it solves is real and nothing on `main` addresses it.

What this release touched, so you know what to expect: `CHANGELOG.md`, `README.md` (the generated test-count badge again, sorry — take whatever `main` has for that line), `docs/user-guide/scanning.md` and `docs/architecture.md`. It did **not** touch `app/services/audiobookshelf.py`, `app/routers/sync.py`, or `app/templates/settings.html`, so your actual change should rebase cleanly this time; the conflict should once more be confined to the badge line.

I am sorry this keeps happening to your branch. It is a known cost of how fast this repo releases, and it is written up in `CONTRIBUTING.md` under "How PRs get merged here" — it is not a judgement about the work.

dgahagan:
Heads-up: [v0.33.0](https://github.com/dgahagan/shelf/releases/tag/v0.33.0) just went out, so `main` has moved again. Rebase onto that tag.

The decision is unchanged from my earlier comments — this is still the one open PR I want and am asking you to rebase rather than close. `get_playback_url` still builds a user-clickable `href` from the same `abs_url` used for server-side API calls, and nothing on `main` addresses it.

What this release touched, so you know what to expect:

- `CHANGELOG.md` — new `## [0.33.0]` section at the top.
- `README.md` — the generated test-count badge line again (2587 → 2596). Sorry. Take whatever `main` has; `make badges` regenerates it and the check does not fail pull-request builds.
- `app/routers/items_csv.py`, `tests/test_csv_roundtrip.py`, `docs/user-guide/import-and-export.md`, `docs/faq.md`, `GOTCHAS.md`.

**None of that overlaps your diff** except the README badge line. Your `app/services/audiobookshelf.py`, `app/routers/sync.py`, `app/templates/` and `static/` changes are untouched by this release, so the rebase should be that one line plus whatever `make css` and the service-worker stamp regenerate.

Also worth knowing: the two CSV fixes in this release came out of your closed #70. I said I would land them and credit you, and both changelog entries do.

dgahagan:
Heads-up: [v0.33.1](https://github.com/dgahagan/shelf/releases/tag/v0.33.1)
just went out, so `main` has moved again. Rebase onto that tag.

The decision is unchanged from my earlier comments — this is still the one open
PR I want and am asking you to rebase rather than close. Nothing in 0.33.1
touches `app/services/audiobookshelf.py`, `app/routers/sync.py` or the settings
templates your branch changes; it is a concurrency fix in the four add paths
(`intake.py`, `items_catalog.py`, `hardcover.py`). So the conflict should still
be the generated files only — `README.md`'s test-count badge line and
`static/css/app.css` / `static/sw.js`. Those are all `make badges` / `make css`
output: take `main`'s side and regenerate rather than hand-merging them.

dgahagan:
Shipped in [v0.34.0](https://github.com/dgahagan/shelf/releases/tag/v0.34.0). Thank you — this is the whole of that release.

I am closing this PR rather than merging it, because the branch had been outrun by four releases and gone conflicting through no fault of yours. I rebased your work onto current `main` myself and kept your authorship: the release commit on `main` is [a994eab](https://github.com/dgahagan/shelf/commit/a994eab), authored by you.

Two changes from the submitted diff, both from my review comment above:

- **The field folds into the existing Audiobookshelf Sync form** and `_INTEGRATION_KEYS`, rather than a dedicated settings card plus a `POST /api/sync/audiobookshelf/public-url` route. It now sits directly under the URL it modifies, sharing one Save button — a user setting the browser URL is already looking at the API URL, so splitting them into two cards asked them to hold both in their head.
- **`get_playback_url` stays a pure string builder** and takes the browser root as an argument. Reading the setting inside it opened a nested connection per call, and one caller maps it over every linked item on the page.

I also added an `ABS_PUBLIC_URL` env override to mirror `ABS_URL`, and moved the URL validator from `app/routers/sync.py` into `app/services/audiobookshelf.py`, since the settings router now validates the same shape — `sync._validate_abs_url` re-exports it.

Your test file came through as written. I test-drove the result against a copy of my real collection (1057 items, 960 with an ABS id) in desktop and mobile viewports: the API URL is never touched by any of it, and clicking **Test** with a deliberately wrong token still gets a 401 from the real server while the browser URL pointed at a host that does not resolve — which is the clearest proof the two URLs are genuinely separate. The env override is confirmed live too.

One thing your PR description called out that I want to acknowledge: splitting this out of #79 so the two reviews stopped overlapping was the right call, and it is why this one could be rebased cleanly at all.

PR #82, opened 2026-09-03
Fix UPC-A check digit validation

martialartistslife:
## Summary
Fix UPC-A check-digit validation by applying the standard weighting:
- positions 1, 3, 5, ... are weighted by 3;
- positions 2, 4, 6, ... are weighted by 1.
The previous implementation applied those weights in reverse, causing valid retail UPC-A values such as `036000291452` and `078073003501` to be rejected.
## Tests
Added focused regression vectors covering:
- valid UPC-A values: `012345678905`, `036000291452`, `078073003501`;
- invalid check digits: `012345678900`, `036000291453`, `078073003502`.
Verification completed before opening this PR:
- focused UPC/route tests: 177 passed;
- static checks passed: CSRF, Alpine CSP, service-worker version, badges, UTF-8 convention lint, and whitespace;
- full unit suite: 2449 passed with 14 unrelated Windows/environment failures involving archive file locks, POSIX permissions, and CP1252 decoding.
No migrations, data changes, or unrelated functionality are included.
Requested by @dgahagan in martialartistslife/shelf#9.

### DIFF
--- a/app/services/upc.py
+++ b/app/services/upc.py
@@ -40,10 +40,15 @@ def detect_barcode_type(code: str) -> str:
 
 
 def validate_upc(code: str) -> bool:
-    """Validate a UPC-A (12-digit) check digit."""
+    """Validate a UPC-A (12-digit) check digit.
+
+    UPC-A weights positions 1, 3, 5, ... by three and positions 2, 4, 6, ...
+    by one. The previous implementation had those weights reversed, which
+    rejected valid retail barcodes such as 078073003501.
+    """
     code = normalize_barcode(code)
     if len(code) != 12 or not code.isdigit():
         return False
-    total = sum(int(d) * (3 if i % 2 else 1) for i, d in enumerate(code[:11]))
+    total = sum(int(d) * (3 if i % 2 == 0 else 1) for i, d in enumerate(code[:11]))
     check = (10 - (total % 10)) % 10
     return int(code[11]) == check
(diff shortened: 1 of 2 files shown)


dgahagan:
Merged — thank you. The reversed weighting was a real bug: every UPC-A whose two weighted sums differ was being rejected, which is most of them. Your three valid vectors (`012345678905`, `036000291452`, `078073003501`) plus the three bad-check-digit ones are exactly the shape this needed — the fix and the regressions landed together, so the next person to touch `upc.py` cannot silently flip the weights back.

This will ship in the next release with the changelog crediting you.

dgahagan:
Shipped in [v0.32.0](https://github.com/dgahagan/shelf/releases/tag/v0.32.0). Thank you — the weights were genuinely reversed and your fix is correct.

I did change your changelog wording, and you should know why rather than find it changed. Your entry said the reversed weights meant valid barcodes "were rejected". While preparing the release I checked, and `validate_upc` had **no caller anywhere in `app/`** before this release — `git grep validate_upc` against the 0.31.0 tree is empty. So no user's scan was ever rejected by it, and as written the entry promised a behaviour change nobody could reproduce.

The entry now describes it as a correctness fix to a validator whose first consumer is #88's legacy barcode support, which lands in this same release and genuinely depends on the answer being right. Your credit and PR link are unchanged.

That makes the fix more load-bearing than it looked, not less: #88 calls `validate_upc` on every legacy barcode it parses, and with the old weights it would have rejected `078073003501` — the exact Scholastic price-point code the feature exists to read. Your two PRs fit together.

PR #84, opened 2026-09-03
Fix infinite scroll appending unstyled grid cards in list view

mattbasta:
## The bug

On the Browse page in **list view**, infinite scroll breaks: the second page of results is appended as unstyled markup below the table, and pagination then stalls entirely. Grid view is unaffected.

Reproduced at a 390px viewport with 150 seeded items — the page rendered 60 rows, appended 60 `<a>` cover cards directly under the table's wrapper `div`, and stopped at 120 of 150.

## Root cause

Three compounding defects:

1. **The view mode never reached the load-more request.** `/browse` builds `load_more_url` server-side (`app/routers/pages.py`), but `viewMode` lives in `localStorage`, so the server can't know it. The first load-more after any page load hit `/api/search?page=2` with no `view`, which the endpoint defaults to grid. `app/routers/items.py` only re-appended `view` when it was non-empty, so the omission propagated to every subsequent page.

2. **The list-view sentinel sat outside the table.** `item_grid.html` placed the list `#load-more` after `</table>`, even though page 2+ returns bare `<tr>` rows meant for `<tbody>`.

3. **The wrong element was being swapped.** `item_rows_page.html` put `id="load-more"` on the `<tr>` but `hx-get`/`hx-swap="outerHTML"` on the `<td>`, so a swap replaced the *cell* and nested `<tr>` inside `<tr>`.

## The fix

- Load-more elements now carry `hx-include="[name='view']"`, sending the live view mode from the hidden input `browse.html` already renders. The redundant URL param is dropped from `items.py` so the two sources can't disagree.
- The list sentinel moved into `<tbody>` as a proper `<tr>`, with the htmx attributes on the row itself.

## Verification

- Both view modes now load all 150 items; rows land inside `<tbody>`, with no stray nodes and no leftover sentinel.
- 467 unit tests, 36 e2e tests, and both `check_alpine_csp` / `check_csrf_fetch` lints pass.
- Added a parametrized e2e regression test covering both view modes at a mobile viewport. Confirmed it fails on the pre-fix code (`only 60 rows loaded`) by reverting the source changes and re-running it.

## Notes for reviewers

The fix relies on the hidden `<input name="view">` in `browse.html` being present. If Alpine fails to initialize, `hx-include` contributes nothing and the request falls back to grid — the previous behavior, so no new failure mode.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/routers/items.py
+++ b/app/routers/items.py
@@ -801,8 +801,9 @@ async def search_items(
     if tag:
         from urllib.parse import quote
         qs_parts.append(f"tag={quote(tag)}")
-    if view:
-        qs_parts.append(f"view={view}")
+    # NOTE: `view` is deliberately NOT baked into the URL — the load-more
+    # element sends it via hx-include="[name='view']" so the fragment always
+    # matches the view mode currently rendered on the client.
     qs_parts.append(f"page={page + 1}")
     load_more_url = "/api/search?" + "&".join(qs_parts)
 
--- a/app/templates/fragments/item_cards_page.html
+++ b/app/templates/fragments/item_cards_page.html
@@ -3,7 +3,8 @@
 {% endfor %}
 {% if has_more is defined and has_more %}
 <div id="load-more" class="col-span-full text-center py-4"
-     hx-get="{{ load_more_url }}" hx-trigger="revealed" hx-swap="outerHTML">
+     hx-get="{{ load_more_url }}" hx-trigger="revealed" hx-swap="outerHTML"
+     hx-include="[name='view']">
     <div class="inline-block w-6 h-6 border-2 border-shelf-border border-t-shelf-accent rounded-full animate-spin"></div>
 </div>
 {% endif %}
--- a/app/templates/fragments/item_grid.html
+++ b/app/templates/fragments/item_grid.html
@@ -6,7 +6,8 @@
     {% endfor %}
     {% if has_more is defined and has_more %}
     <div id="load-more" class="col-span-full text-center py-4"
-         hx-get="{{ load_more_url }}" hx-trigger="revealed" hx-swap="outerHTML">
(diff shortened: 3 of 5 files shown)


dgahagan:
Thank you for this — the writeup is excellent. A 390px viewport, 150 seeded items, exact counts (60 rows, 60 appended cover cards, stalling at 120 of 150) and a three-part root cause is a better bug report than most bug reports, let alone a PR.

The awkward news is that all three defects are already fixed on `main`, and you had no way to see that.

| Your finding | Where it landed |
|---|---|
| View mode never reaches the load-more request | `3fe5551` — `view` is now a registry property, `app/browse_filters.py:181` (`in_url=False`), and all four sentinels carry `hx-include="[name='view']"` |
| List sentinel outside `</table>` | `7e70c9c` — it is a `<tr>` inside `<tbody>` |
| `id` on the `<tr>` but `hx-get`/`hx-swap` on the `<td>` | `d941260` — all four attributes are on the `<tr>` |

Your two proposed E2E tests also already exist, as `tests/e2e/test_browse.py:177` and `:204`.

You branched from `a99b19a`, which is the 0.30.0 tag — **yesterday's release**. That is the part I want to own: this project has been cutting up to three releases a day, each landing here as one squashed commit rewriting 20+ files, and nothing in the repo told you that or showed you what was in flight. You did everything right and still wrote a PR against a bug fixed three weeks earlier.

I have written that down: the new [CONTRIBUTING.md](https://github.com/dgahagan/shelf/blob/main/CONTRIBUTING.md) now explains the cadence, says plainly that a stale PR's conflict is its branch point being overwritten rather than anything wrong with the code, and commits me to giving every open PR a decision at each release. I would rather lose a PR to "already fixed" than have someone spend an evening the way you just did.

Closing as obsolete — a rebase would produce an empty diff. Please do send the next one; a report at this standard against current `main` is exactly what I want.

PR #85, opened 2026-09-03
Fetch the Open Library work record once per ISBN lookup

mattbasta:
## What this changes

`openlibrary.lookup()` fetches the work record once and shares it between the
author chain and the description, instead of fetching the same document twice.
An ISBN lookup makes three Open Library requests instead of four.

## Why

Adding a book by ISBN made four sequential requests, each behind the per-host
rate limiter, and two of them were the same work record — once through
`_resolve_author`, once through `_resolve_description` → `get_work_description`.

The comment on the second fetch conceded the duplication:

```python
# We may have already fetched the work in _resolve_author, but keeping it
# simple — the rate limiter handles it
```

The limiter does handle it, by making it strictly slower. It cannot skip a
request that was issued, so the second fetch of a document already in hand
costs a full round trip *plus* a gate. Open Library edition/work requests
routinely take a second or more each, so that is dead time the scan card sits
on with every add.

Behaviour is unchanged, including the contracts the enrichment leg documents:

- A non-200 on the work or author request still returns the edition hit minus
  those fields, rather than laundering it into "no such book" (G47). The work
  fetch sits inside the same `try` as before, so a dead socket there is still
  caught by the enrichment guard.
- An edition with no work of its own still falls back to its own author list,
  without asking for a work at all.
- `get_work_description()` stays a standalone entry point — `services/synopsis.py`
  calls it directly for the backfill — now sharing `_fetch_work` with `lookup`.

One incidental hardening: `_work_key` reads `works[0].get("key")`, so a
malformed `works` entry no longer raises `KeyError` where the old
`works[0]['key']` would have.

## How it was tested

- [x] `make test` passes — 2495 passed. The one failure is
      `test_badge_stamp.py::test_readme_badges_are_current`, which is the
      expected local result for a PR that adds tests; it is `skipif`-ed on
      `GITHUB_EVENT_NAME == pull_request`, and I have deliberately **not**
      restamped the badge, per that test's own docstring about restamps
      colliding across a batch of PRs. Verified green under a PR context:
      `GITHUB_EVENT_NAME=pull_request pytest tests/test_badge_stamp.py` →
      5 passed, 1 skipped.
- [ ] `make test-e2e` — not run.
- [ ] `make checks` — not run in full. The change is one service module with no
      new dependencies, no template or JS changes, and no migration, so the
      CSRF/Alpine/CSS gates are untouched; `test_csrf_lint` and
      `test_alpine_csp_lint` pass as part of the suite above.
- [ ] `make css` — not applicable, no templates or Tailwind classes touched.

Two tests added to `TestOpenLibraryOutcomes` in `tests/test_outbound_clients.py`,
alongside the existing enrichment-failure test:

- `test_the_work_is_fetched_once_for_author_and_description` asserts
  `work.call_count == 1` and a total of three calls, so the redundant fetch
  cannot come back unnoticed. Confirmed non-vacuous: it fails `assert 2 == 1`
  against `main` as it stands.
- `test_an_edition_with_no_work_asks_for_no_work` covers the other side of
  sharing one fetch.

## Notes for review

- Deliberately left out: the same request path also does an extra Hardcover
  round trip when a token is configured. `_lookup_metadata`'s "enrich if series
  or description is missing" leg fires on essentially every add, because Open
  Library never carries series data — so it is a guaranteed serial lookup
  waiting out the whole Open Library chain, and the Google-Books fallback path
  issues two full Hardcover lookups. Running it concurrently with the primary
  leg looks like a clear win, but it touches the `ProviderResult` cascade's
  `legs` ordering and the national-provider short-circuit, which have careful
  documented semantics. That felt like it wanted its own issue and its own PR
  rather than being folded in here. Happy to open one if you'd like it.
- Also on that path: both `/api/scan` and `/api/books/add` download the cover
  inline before responding. `covers.download_cover`'s first candidate needs
  only `cover_id`, which arrives with the *edition* record, so it could overlap
  the work/author requests — but that needs `lookup` split into two phases and
  `download_cover` decoupled from `item_id`. Noted, not attempted.
- I have not exercised this against the live Open Library API, only against the
  respx mocks in the suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to Shelf are documented here. The format follows
 
 ## [Unreleased]
 
+### Changed
+
+- **Adding a book by ISBN makes one fewer Open Library request.** The work
+  record backs both the author chain and the description, and each resolver
+  was fetching it separately — the same document twice, a round trip and a
+  rate limiter gate apart. Open Library edition and work requests routinely
+  take a second or more each, so scanning now settles noticeably sooner. The
+  result is unchanged, including a failed work or author request still
+  returning the edition it already had.
+
 ### Fixed
 
 - UPC-A check digits are validated with the standard weighting (odd positions
--- a/app/services/openlibrary.py
+++ b/app/services/openlibrary.py
@@ -98,11 +98,17 @@ async def lookup(isbn: str, client: httpx.AsyncClient) -> provider_result.Provid
     # (G47). The edition is already a hit — a failed enrichment costs fields,
     # not the result.
     try:
-        author = await _resolve_author(data, client)
+        # The work record backs both the author chain and the description, so
+        # fetch it once and share it. Two requests for the same document cost
+        # a round trip plus a rate limiter gate on every ISBN add.
+        work_key = _work_key(data)
+        work = await _fetch_work(work_key, client) if work_key else None
+
+        author = await _resolve_author(data, work, client)
         if author:
             result["authors"] = author
 
-        desc = await _resolve_description(data, client)
+        desc = _work_description(work)
         if desc:
             result["description"] = desc
     except Exception:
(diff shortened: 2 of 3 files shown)


dgahagan:
Thanks — the change itself is right, and I want it. Collapsing four sequential Open Library requests to three matters more than the count suggests, because each one waits on the per-host pacer in `outbound.py`, so the saved request is saved wall-clock on every ISBN add.

One thing before I can merge: it now conflicts, but **only on `CHANGELOG.md`** — 0.31.0 was cut earlier today and took the lines your entry sits on. `app/services/openlibrary.py` and `tests/test_outbound_clients.py` still apply cleanly. If you rebase on `main` and move the entry under `## [Unreleased]`, this goes in as-is; CI was already green on the code.

mattbasta:
Rebased onto current `main` (`f44f0dd`) — thanks, and the timing was instructive: the conflict was exactly the shape your new CONTRIBUTING section describes. 0.31.0 consumed the `[Unreleased]` section my entry was sitting in, so the only conflicting file was `CHANGELOG.md`; the code hunks applied clean.

Resolution keeps both entries under `[Unreleased]`, with `### Changed` ahead of the UPC `### Fixed` from #82 to match Keep a Changelog's ordering. No generated files touched — no `app.css`, no `SW_VERSION`, and I have deliberately left the README test-count badges alone.

Re-ran the suite on the new base: 2559 passed, with `test_badge_stamp.py::test_readme_badges_are_current` the only local failure, which is the expected result for a PR that adds tests and is `skipif`-ed on `pull_request`. CI on the rebased head is green across `test`, `css` and `e2e`.

dgahagan:
Shipped in [v0.32.0](https://github.com/dgahagan/shelf/releases/tag/v0.32.0). Thank you.

Verified live during the release test drive rather than only in the suite: scanning ISBN `9780441013593` against a real instance filed *Dune* with the author chain and description both intact, so folding the two work fetches into one did not cost either resolver its data — which was the thing worth checking, since they were fetching the same document independently.

The changelog entry is under `### Changed` and credits you by handle and PR number.

One small correction on my side, for the record: your PR was green on its own branch but the merge to `main` went red immediately afterwards, on the README test-count badge. That was not your doing — the badge check is deliberately skipped on `pull_request` builds and enforced only on push to `main`, precisely so a batch of PRs does not become mutually unmergeable on one generated line. It is the maintainer's job to restamp after the merge, and I had not. Fixed in the release.

PR #88, opened 2026-09-03
Add safe legacy UPC+5 book barcode support

martialartistslife:
## Summary

Support the evidence-backed Scholastic price-point UPC-A + 5 book barcode format without treating the shared UPC as an ordinary product barcode.

- Recognize the supported 078073 format in its canonical 17-digit form and exact leading-zero 18-digit scanner representation.
- Validate the UPC-A check digit and reject unsupported prefixes, arbitrary EAN-13+5 values, and invalid barcodes.
- Generate only the finite checksum-valid ISBN candidates implied by the known Scholastic prefixes, then verify every candidate through Shelf's existing metadata cascade.
- Resolve exactly one verified candidate automatically; keep multiple verified candidates ambiguous and fail closed when provider results are uncertain.
- Present a deliberate title choice for ambiguity and remember the selected ISBN only for that exact barcode identity, with validation on reuse.
- Cover add, wishlist, lend, return, move, inventory, lookup, and quick-rate modes, including camera-inventory accounting after an ambiguity choice.
- Protect the submitted choice from provider-result races.

Real acceptance examples include:

- 07807300350143506 / 007807300350143506 → 059043506X / 9780590435062
- 07807300399044891 → an explicit choice for Hound at the Hospital (9780439448918)

## Data compatibility

Adds append-only migration 24 for confirmed barcode mappings and defines the same table on the fresh-database path. The migration is non-destructive; no existing item data is rewritten.

## Tests

- Focused legacy-book, route, schema-parity, and scan-mode tests: 91 passed
- Complete scan Playwright file: 50 passed
- CSRF, Alpine CSP, test-convention, badge, service-worker, whitespace, and staged secret checks passed
- Dependency audit: no known vulnerabilities
- The full local non-E2E run completed 2,575 tests successfully; the remaining failures are Windows/environment-only archive file locks, POSIX permission expectations, and existing path-separator/allowlist convention checks. The generated service-worker stamp was then corrected and its dedicated tests pass.

Requested by @dgahagan in martialartistslife/shelf#9, based on the final states of martialartistslife/shelf#8 and #9.


dgahagan:
This is good work, and the safety model is the best part of it — I want it in. Two suggestions, both written out on [`suggest/pr88-fixes`](https://github.com/dgahagan/shelf/compare/suggest/pr88-fixes) so you can look at real commits instead of a description. They sit directly on your `c08d79b`; cherry-pick, squash into your branch, or push back on either.

**I tried a third change and your tests correctly stopped me.** Worth saying first, because it is the most useful thing I learned. Since `resolve` costs a cascade per candidate, I added a shortcut: if exactly one candidate ISBN is already an item in the collection, take it and skip the providers. Two of your tests failed immediately — `test_ambiguous_barcode_cannot_mutate_any_existing_item_mode` and `test_inventory_confirmation_marks_the_item_for_camera_accounting`. Both seed exactly one owned candidate and assert the scan is *still* ambiguous. You are right and I was wrong: owning the `0590` original is not evidence that the copy in your hand is not the `0439` reprint, and my shortcut would have filed the wrong one silently — the exact failure the module exists to prevent. I backed it out. Thank you for writing those tests that way.

**1. One ISBN-10 check-digit derivation** (`021faad`). `legacy_book._isbn10_check_digit` is arithmetically identical to the tail of `isbn.isbn13_to_isbn10` — same descending 10..2 weighting, same modulus 11, same `X` at 10. I extracted `isbn.isbn10_check_digit` and pointed both at it. The reason I care more than usual here: #82, merged the same day, was a check-digit bug caused by duplicated weighting written with the factors reversed. A second private copy is how that recurs somewhere the first fix cannot reach. The new test asserts both derivations agree and that the private helper does not come back.

**2. Remember an automatic resolution, not only a chosen one** (`f869c73`). `_save_confirmed_legacy_mapping` has one call site, inside the `confirmation_accepted` branch, so only an ambiguity resolved by hand is ever remembered. The `elif resolution.outcome == "found"` branch below it does not save, which means an unambiguous legacy barcode re-runs the whole resolver on every scan, forever.

That is expensive in a way the code does not show, and it is a consequence of a design decision I think is right: `resolve` deliberately cannot short-circuit, because it has to see every candidate to rule out ambiguity. So every scan pays a full `_lookup_metadata` cascade *per candidate* — national, Open Library, Hardcover, Google Books, all paced through `outbound.py`. In inventory mode that is per book, per sweep. The new test pins it: two cascades on the first scan, zero on the second.

My argument that caching `found` is safe is your own: `found` already means one candidate verified positively and every other returned an ordinary miss, because a candidate that could not be checked yields `inconclusive` instead and never reaches that branch. So it stores the same evidence a title choice stores, minus the choosing. Tests assert neither an `inconclusive` nor a `not_found` scan is remembered.

The trade, documented at the call site: a candidate that misses today because a provider lacks it, but gains a record later, stays ruled out for that barcode until the row is cleared. If you think that should re-verify periodically instead, say so — you have thought about this failure mode more carefully than the diff required, and I would rather take your read.

**Two things I will handle at merge, not you:**

- `CHANGELOG.md` needs the credit line the #82 entry above it carries — `Contributed by @martialartistslife in #88`. My convention, my omission to fix.
- The README test-count badge is restamped in your diff (2585). Please leave badges alone in future PRs — they are generated by `make badges` and go stale the moment anything else merges, which is exactly what happened when #85 landed two tests. The check no longer fails pull-request builds, so a stale badge costs you nothing. I will regenerate it.

Verified locally on top of your head: `2592 passed`, with only `test_readme_badges_are_current` failing, which is the expected local result for a PR that adds tests and is skipped under `pull_request`. `make checks-fast` is clean, including `check-sw-version` — your generated assets are consistent.

Smaller notes, none blocking: the `htmx:afterSwap` fallback in `scan.js` reads as belt-and-braces but is actually the only path that fires, since `e.detail.target` is the detached old element for an `outerHTML` swap — worth a comment saying so. The two `assert` statements in `scan_isbn` are inert under `python -O`; both are genuinely unreachable, so it is style only. And a valid UPC with an unsupported publisher prefix ends up reported as "Invalid ISBN", which is safe but misleading for something Shelf can see is a price-point barcode.

dgahagan:
Shipped in [v0.32.0](https://github.com/dgahagan/shelf/releases/tag/v0.32.0). Thank you — this is a careful piece of work and it held up under a live test drive on a real 1000-item collection.

Two things worth telling you, because they say something your tests could not.

**The ambiguity you refused to resolve is real.** Driving `07807300350143506` against live Open Library returned two genuinely different published books — "Bsc Kristy and the Mothers Day Surpris" (Ann M. Martin, `059043506X`) and "Merry Christmas Hungry Bear!" (Don Wood, `0439435064`). Had the resolver picked one, it would have filed the wrong book roughly half the time. The decision to stop and ask is doing real work, not guarding a hypothetical, and the same is true of keeping a provider failure out of the "no match" bucket.

**Migration 24 landed cleanly on a database that needed it most.** The test copy had no `schema_version` table at all and was missing `items.language`. It backfilled, created `legacy_book_mappings`, and kept all 1057 items. Appending the migration *and* mirroring it into `SCHEMA` is exactly right (G1), and your two tests pinning append-only order are the reason I did not have to check by hand.

**Where I went a different route:** your changelog entry landed under `### Fixed`; I moved it to `### Added`, since it is a new capability rather than a repair, and expanded it for the release notes. Nothing about the behaviour changed.

**One follow-up I opened rather than asked you to fix:** [#90](https://github.com/dgahagan/shelf/issues/90). Scanning the same book's bare 12-digit UPC — no supplement — still files a DVD, because `parse()` needs 17 digits and a 12-digit read never reaches your code. That is pre-existing and not yours to fix, but your PR is what made it visible, and it matters because plenty of scanners need configuring before they read a five-digit add-on. The fix is constrained in a way worth knowing: Shelf deliberately files a placeholder rather than rejecting an unreadable barcode (`CONTRIBUTING.md:63`), so "just refuse the scan" would trade a wrong row for a lost one.

**One nit, not worth a PR:** `resolve()`'s docstring says a credential failure makes the result inconclusive, but `no_credential` is in the same set as `no_match`. Unreachable in practice — `combine` only reports it when every leg was skipped, and Open Library needs no credential — so it is a doc/code disagreement rather than a live path.

PR #93, opened 2026-09-06
Allow manual cover image URLs

### LINKED ISSUE #91: Allow manual cover image URLs

Shelf's cover picker can search metadata providers or upload a local image, but there is no way to use a specific cover image that already exists online without downloading it first.

This comes up when the correct edition/pressing/issue artwork is available from a publisher, archive, retailer or other public image source but is not returned by Shelf's configured metadata providers. The current workaround is to download the image manually and then upload it to Shelf.

Add a small "Use image from URL" option to the existing cover picker.

The user would paste a public HTTPS image URL and Shelf would save its own local copy, just as it does for provider-selected or uploaded artwork. A failed URL should leave the current cover unchanged.

Because this makes Shelf fetch a user-supplied URL, the server-side implementation should reject private/local network destinations, validate redirects, enforce the existing image type/size limits, and require Editor access.

- Download the image locally and use the existing Upload control.
- Rely on Find cover/provider results even when they do not contain the exact artwork wanted.

I have a small implementation prepared against current 0.34.0 and can submit it as a focused PR if this approach fits Shelf.

sudo-rpaisley:
## What this changes

Adds a **Use image from URL** option to the existing cover picker.

- accepts a user-supplied public HTTPS image URL;
- downloads and stores Shelf's own local copy rather than hotlinking;
- keeps the existing cover unchanged if the download or image validation fails;
- requires Editor access, matching the existing cover controls;
- rejects private/local/mixed DNS destinations;
- pins the HTTPS connection to the validated public address while preserving the original hostname for TLS/SNI and Host validation;
- validates every redirect again, with a five-redirect limit;
- streams responses under Shelf's existing 10 MB cover ceiling and existing JPEG/PNG/GIF/WebP validation.

Fixes #91.

## Why

The correct cover for a particular edition, pressing or issue is sometimes available online but not returned by Shelf's metadata providers. Today the image has to be downloaded manually and then uploaded to Shelf.

This keeps that workflow inside the existing cover picker while treating a pasted URL as untrusted network input.

## How it was tested

GitHub Actions validation run #829 passed on the exact contribution branch:

- [x] `make test` passes
- [x] `make test-e2e` passes
- [ ] `make checks` passes

The CI offline secret/CSRF/Alpine checks also pass. I have not marked the full `make checks` target because the validation workflow runs its offline lint subset rather than the network-dependent dependency audit.

Focused tests cover public/private DNS handling, mixed answers, redirect revalidation, pinned TLS/Host behaviour, content type/signature and size limits, failed-download preservation, Editor permissions and successful local persistence.

## Notes for review

This is deliberately isolated from the larger fork: no schema changes, no library/ACL dependencies and no unrelated editor work.

The manual-URL path does not reuse provider cover downloading directly because provider URLs are trusted application inputs whereas this endpoint accepts arbitrary user-supplied network destinations. The separate helper keeps that SSRF boundary explicit.

### DIFF
--- a/app/routers/items_covers.py
+++ b/app/routers/items_covers.py
@@ -18,7 +18,7 @@ from app.auth import require_role
 from app.config import COVERS_DIR, HTTP_TIMEOUT
 from app.database import get_db, get_setting
 from app.routers import items_common
-from app.services import covers, cover_queue, openlibrary, scan_outcome
+from app.services import covers, cover_queue, manual_cover, openlibrary, scan_outcome
 from app.services import isbn as isbn_svc
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


dgahagan:
Shipped in [v0.35.0](https://github.com/dgahagan/shelf/releases/tag/v0.35.0) — `docker pull dangahagan/shelf:0.35.0`. Thank you.

I want to call out the security work specifically, because it is the part of this PR that took the real thought and it is the part a reviewer would most easily have waved through. Fetching a URL the user supplies is a server-side request forgery hole by default, and you closed it properly rather than settling for a scheme check: resolving off the event loop, requiring **every** returned address to be globally routable — rejecting a mixed public/private answer outright instead of filtering it, which is what actually closes DNS rebinding — then pinning the validated IP for the connection while keeping the hostname for the Host header and SNI. Re-resolving and re-validating each redirect hop rather than trusting the first one is the detail most implementations get wrong. The size cap enforced twice, against `Content-Length` and again while streaming, is the right shape too.

The uniform failure return is a good call and I kept it exactly as you wrote it: every rejection reason collapses to the same `None` and the same message, so the endpoint cannot be used to probe what is reachable from the server. That is worth more than per-reason error text would have been.

**Where I added to it rather than changed it:** I documented the path in `docs/architecture.md` as the one outbound fetch that deliberately does *not* go through `services/outbound.py`, and said why — the host is user-supplied, so there is no published rate limit to pace against and no allow-list entry to match, and what it needs instead is containment the paced clients have no reason to carry. That was a structural fact with no user-visible symptom, so it would have rotted silently if it stayed only in the code. I also wrote the user-facing half into `docs/user-guide/items.md` and pointed `docs/troubleshooting.md` at it from the two cover-miss cases where it is the actual remedy: a 979-12 ISBN, and a DVD or game with no TMDb/IGDB match.

No code changes from me — it went out as you wrote it.

PR #94, opened 2026-09-06
Scan ISBNs from the item edit page

### LINKED ISSUE #92: Scan ISBNs from the item edit page

Shelf can already scan ISBNs when adding items, but correcting or replacing an ISBN on an existing item's Edit page requires typing the number manually. On a phone this is slower and more error-prone than scanning the barcode that is already on the book.

This is especially useful when correcting an older catalogue record, replacing an incorrect identifier, or adding an ISBN that was missing when the item was first entered.

Add a small **Scan ISBN** action beside the existing ISBN field on the normal item Edit page.

Pressing it would open Shelf's existing camera barcode scanner and accept a 13-digit Bookland EAN beginning `978` or `979`. A successful scan would only place the value into the ISBN field and close the camera.

It should **not** save the item or perform a metadata lookup automatically. The user can review or edit the scanned value and then press Save normally, leaving Shelf's existing server-side ISBN validation unchanged.

The implementation can reuse Shelf's already-vendored html5-qrcode / ZXing scanner stack and the existing compact camera framing.

- Type the ISBN manually from the book.
- Leave the Edit page, use the normal Add/Scan workflow, then reconcile the existing catalogue entry.
- Copy/paste the ISBN from another source.

I have a focused implementation prepared directly against current 0.34.0 with no route, schema, or database changes. It is isolated to the existing edit-page JavaScript plus regression coverage.

sudo-rpaisley:
## What this changes

Adds a **Scan ISBN** action for the ISBN field on the normal item Edit page.

- lazily loads Shelf's existing vendored html5-qrcode / ZXing scanner stack when the camera is requested;
- uses the same compact barcode framing as the existing scan page;
- accepts only 13-digit Bookland EAN values beginning `978` or `979`;
- places the decoded ISBN into the existing edit field and returns focus to it;
- does **not** submit the form or perform a metadata lookup automatically;
- leaves the existing server-side ISBN validation unchanged and authoritative when Save is pressed.

Fixes #92.

## Why

Correcting or adding an ISBN on an existing item currently requires typing or pasting it, even though Shelf already has a camera barcode scanner elsewhere. On a phone, scanning the barcode is faster and avoids transcription mistakes while still letting the user review the value before saving.

## How it was tested

The focused regression verifies shared scanner reuse, 978/979 filtering, compact framing and the no-auto-submit/no-fetch contract.

- [x] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

GitHub Actions passes the unit/integration suite and offline secret/CSRF/Alpine checks on the final one-commit branch. A previous run of the same scanner code passed the scanner-related browser coverage; a later full-suite run had one unrelated failure in the existing bulk-move E2E test. The fresh exact-head E2E run is still in progress.

I have not marked the full `make checks` target because the validation workflow runs its offline lint subset rather than the network-dependent dependency audit.

## Notes for review

This is intentionally narrower than the fork's broader barcode-edit work: no UPC/EAN persistence changes, magazine supplements, new media types, database changes or route overrides are included.

The scanner only fills the existing ISBN input. All normal edit semantics and validation remain unchanged.

### DIFF
--- a/app/templates/item_edit.html
+++ b/app/templates/item_edit.html
@@ -33,7 +33,7 @@
     </div>
     {% elif error == "unknown_platform" %}
     <div data-testid="edit-error" role="alert" class="mb-4 rounded-lg border border-shelf-error/40 bg-shelf-error/10 px-4 py-3 text-sm text-shelf-error">
-        That game platform isn't in your platform list — pick one, or add it under Settings first. Nothing was saved.
+        That game platform isn't in your platform list — pick one from the list. Nothing was saved.
     </div>
     {% elif error == "invalid_reading_status" %}
     <div data-testid="edit-error" role="alert" class="mb-4 rounded-lg border border-shelf-error/40 bg-shelf-error/10 px-4 py-3 text-sm text-shelf-error">
@@ -185,17 +185,11 @@
                 <h2 class="text-lg font-semibold">Identifiers</h2>
                 <p class="text-sm text-shelf-muted mt-1">Identifiers printed on this edition.</p>
             </div>
-            <div>
-                <label for="isbn" class="block text-sm font-medium text-shelf-muted mb-1">ISBN</label>
-                <div class="flex flex-col sm:flex-row gap-2">
-                    <input type="text" id="isbn" name="isbn" value="{{ item.isbn if item.isbn is not none else '' }}" placeholder="ISBN-13"
-                           class="flex-1 bg-shelf-bg border border-shelf-border rounded-lg px-3 py-2 text-shelf-text focus:ring-2 focus:ring-shelf-accent focus:border-transparent outline-none">
-                    <button type="button" @click="startCamera()"
-                            class="px-4 py-2 bg-shelf-hover text-shelf-text border border-shelf-border rounded-lg text-sm hover:border-shelf-accent/50 transition-colors">
-                        Scan ISBN
-                    </button>
-                </div>
-            </div>
+            {{ field("isbn", "ISBN", item.isbn, placeholder="ISBN-13") }}
+            <button type="button" @click="startCamera()"
+                    class="px-4 py-2 bg-shelf-hover text-shelf-text border border-shelf-border rounded-lg text-sm hover:border-shelf-accent/50 transition-colors">
+                Scan ISBN
+            </button>
 
             <div x-show="cameraActive" style="display:none"
                  class="fixed inset-0 z-50 bg-black/80 p-4 flex items-center justify-center">
--- a/app/templates/item_edit.html
+++ b/app/templates/item_edit.html
@@ -1,5 +1,9 @@
 {% extends "base.html" %}
 {% block title %}Edit {{ item.title }} — Shelf{% endblock %}
+{% block head %}
+<script src="/static/vendor/html5-qrcode-2.3.8.min.js"></script>
+<script src="/static/vendor/zxing-browser-0.1.5.min.js"></script>
+{% endblock %}
 {% block content %}
 <div class="max-w-4xl mx-auto" data-item-edit-sections>
     <a href="/item/{{ item.id }}{% if back.key %}?from={{ back.key }}{% endif %}" class="text-shelf-muted hover:text-shelf-text text-sm mb-4 inline-block">&larr; Back to item</a>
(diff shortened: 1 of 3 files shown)


dgahagan:
Deferring this one to the next release, and I want to be clear that it is not a judgement on the change — it reads well and the checks are green.

#96 (Organise item editing into clear sections) just merged, and it inserts into `static/js/item_edit.js` at the same anchor this PR does, immediately after `coverDrop()`. The two are independent in substance but GitHub cannot auto-resolve the overlap, and `maintainerCanModify` is off on this branch so I cannot rebase it for you.

The rebase is worth doing on its own merits rather than just to clear the conflict: #96 gives the edit page an **Identifiers** section, and that is where a Scan ISBN control belongs — sited inside that section rather than appended after the input. If you rebase onto `main` and move the button into the Identifiers section, I will take it in the next release.

One thing to check when you do: 0.35.0 (going out today) adds a script load-order lint (`make check-alpine`) and a component-load guard with a closed declaration map in `static/js/component-load-guard.js`. Your lazy `loadLocalScript` injection is exactly the kind of thing the new lint has an opinion about, so run `make check-alpine` after rebasing — your current green CI ran against pre-0.35.0 `main` and has not seen it.

Thanks for the batch — #93 and #96 are both in 0.35.0.

dgahagan:
Thanks for this — holding it until after the 0.35.0 release rather than merging it now, and here's exactly why.

Your own #96 ("Organise item editing into clear sections") landed a few hours ago and rewrote `static/js/item_edit.js`, which is the file this PR builds on. GitHub now reports #94 as CONFLICTING, so it can't be squash-merged, and its CI can't run either — GitHub can't build a merge commit for it. That's a collision between two of your own PRs on the same file, not a problem with this one's substance.

**What I'd ask for:** a rebase onto `main` once 0.35.0 is tagged (later today). The conflict itself should be mechanical — #96 added `updateEditSectionVisibility` plus a `DOMContentLoaded` listener above the `alpine:init` block, and your scanner code lands in the same region.

**One design note worth raising before you spend the rebase effort**, since it may change what you want to submit: this PR builds the whole scanner UI imperatively in JS — creating the button, overlay, panel, heading, video and status elements with `document.createElement`. Everything else in this app is server-rendered Jinja with Alpine for behaviour, and the existing `/scan` page already owns barcode scanning including the html5-qrcode/ZXing fallback chain. A version that renders the overlay markup in `item_edit.html` and reuses `/scan`'s scanner setup rather than re-implementing it would fit the codebase better and be a lot smaller. Happy to look at either shape — I just didn't want you to rebase 190 lines I'd then ask you to restructure.

Not closing it; it stays open and I'll pick it back up after the rebase.

dgahagan:
Merged — thank you for the restructure, this is exactly the shape I was hoping for. The overlay markup in the Identifiers section, `window.createBarcodeScanner` instead of a second scanner, and `{% block head %}` for the two vendored files: all of it fits the way the rest of the app is built, and it came in at 137 lines instead of 190.

The 978/979 filter with the retry message is a nicer answer than silently rejecting, and I like that the regression test pins the *contract* (`.submit(` and `fetch(` both absent from the source) rather than just the happy path. That's the part that will still be protecting this in a year.

**One thing I owe you an apology for.** In your PR description you wrote off "one unrelated failure in the existing bulk-move E2E test" — [run 34049245957](https://github.com/dgahagan/shelf/actions/runs/34049245957), failing on `test_bulk_move_apply_moves_selected_list_item`. You were right that it was unrelated to your change, and you were more right than you knew: that is a genuine race in **our** test, not a flake and not anything you did. The test clicked the bulk-move apply button and then waited on `networkidle` before reading the database, which is a wait on nothing in particular — the assertion could and did run before the POST landed.

I have a branch fixing that class of bug across the whole E2E suite right now (wait on the specific response or navigation the click causes, never on `networkidle`), plus a lint that fails any new `networkidle` wait after a click. It'll be in the next release. Sorry you burned attention on it.

Thanks again.

dgahagan:
Shipped in **[v0.36.0](https://github.com/dgahagan/shelf/releases/tag/v0.36.0)** — thanks, @sudo-rpaisley.

Small PR, good judgement in it. Two decisions I want to name because they are the difference between this being useful and being a trap:

**It fills the field and does not save.** A scanner that wrote straight through to the database would be faster and much worse — the whole reason to scan on the *edit* page is that the stored ISBN is already wrong, and you want to see what the camera read before committing to it. Filling and selecting the input is exactly right.

**It rejects a non-978/979 barcode instead of accepting it.** `acceptDecoded` requiring 13 digits with a 978 or 979 prefix means pointing it at a DVD case says so rather than filing a UPC in the ISBN column. That is precisely the failure mode this app has had reported before, from other directions.

**Verified at both viewports on real data.** The Identifiers section renders one **Scan ISBN** button, sited inside the section rather than trailing after the input. Headless Chromium has no camera, so the drive exercised the failure branch: the red toast reads *"Camera access denied. Check browser permissions for this site."*, the overlay closes, the existing ISBN is untouched, and there are **zero** `pageerror`s. The HTTPS-versus-permission distinction in the error path is a nice touch — a user on plain HTTP gets told the actual problem instead of a permissions dead end.

Documented in [Items → Editing](https://github.com/dgahagan/shelf/blob/main/docs/user-guide/items.md) and cross-referenced from [Scanning](https://github.com/dgahagan/shelf/blob/main/docs/user-guide/scanning.md), which previously implied the Scan tab was the only place a camera scan happens.

PR #96, opened 2026-09-06
Organise item editing into clear sections

### LINKED ISSUE #95: Organise item editing into clear sections

The item Edit page is becoming harder to scan as Shelf supports more media types and more editable metadata. Core metadata, artwork, series, identifiers, location and media-specific details all compete for space in one long form, which makes useful controls easy to miss and is especially awkward on smaller screens.

This becomes more noticeable as new capabilities are added: artwork actions, barcode tools and format-specific metadata all need somewhere predictable to live without turning Edit into an ever-longer list of fields.

Organise the existing item editor into clear, named sections while keeping the current save behaviour and validation model.

A proposed structure is:

- **General** — title, creators, year, description and other core metadata
- **Artwork** — current artwork and the existing search/upload/remove/retry controls
- **Series** — series membership and position
- **Identifiers** — ISBN and other identifiers supported by the item
- **Location** — the item's existing location field
- **Media Details** — fields specific to the selected media type

The sections should be easy to navigate on desktop and mobile, and fields should only appear where they are relevant to that media type.

This should remain one item-editing workflow rather than creating separate save paths. Existing server-side validation, permissions and item-save semantics should remain authoritative.

Where optional features are available, their controls should live in the relevant section rather than adding separate standalone screens. For example, manual artwork URLs would belong in Artwork if #91 lands, and camera ISBN scanning would sit with the ISBN field if #92 lands.

- Keep the current long form and continue adding fields to it.
- Split different kinds of metadata into separate edit pages, which would reduce clutter but make editing one item more fragmented.
- Hide advanced fields behind ad-hoc expanders, which does not provide a consistent information structure.

I have a sectioned editor working in my fork. If this direction fits Shelf, I can prepare a clean contribution from current upstream `main`, preserving the existing edit/save behaviour and excluding fork-specific media types or schema changes.

sudo-rpaisley:
## What this changes

Reorganises the existing item Edit page into six clear sections without changing its save or validation model:

- **General** — core title and descriptive metadata;
- **Artwork** — the existing cover preview and artwork controls;
- **Series** — the existing series membership fields;
- **Identifiers** — ISBN and other identifiers already supported by the item;
- **Location** — the existing location field;
- **Media Details** — fields relevant to the selected media type.

A compact section-navigation strip links to each section using ordinary anchors, so the whole editor remains one form and one save path.

Media-specific fields are shown only when relevant to the selected media type, while legacy populated values remain visible so existing data cannot become inaccessible. Changing Media Type updates the relevant fields immediately.

Fixes #95.

## Why

The edit form has accumulated enough metadata that unrelated fields are increasingly difficult to scan, especially on smaller screens. Giving existing controls predictable homes makes the editor easier to navigate and gives future optional controls somewhere coherent to live without creating separate editing workflows.

This PR is intentionally structural rather than architectural: it reorganises the fields that already exist upstream and does not introduce new media types, schemas or persistence rules.

## How it was tested

The final one-commit contribution branch was validated against current upstream 0.34.0.

- [x] `make test` passes
- [x] `make test-e2e` passes
- [ ] `make checks` passes

GitHub Actions run #837 passed the unit/integration suite, offline secret/CSRF/Alpine CSP checks, CSS reproducibility and the full E2E suite on the exact final head.

Focused regression coverage checks:

- all six sections are present;
- the existing item-edit form action and single Save contract are unchanged;
- every existing upstream edit field remains in the form;
- media-specific field visibility follows the selected media type;
- populated legacy fields remain reachable.

I have not marked the full `make checks` target because the validation workflow runs its offline lint subset rather than the network-dependent dependency audit.

## Notes for review

This is deliberately independent of the other open contributions:

- no manual cover URL support from #93;
- no camera ISBN scanning from #94;
- no fork-only music formats;
- no copies/holdings or hierarchical-location schema;
- no multi-user/library ACL work;
- no integration-specific fields.

That keeps #95 reviewable on its own. If #93 or #94 land later, their controls have natural homes in Artwork and Identifiers respectively without #95 depending on them.

### DIFF
--- a/app/templates/item_edit.html
+++ b/app/templates/item_edit.html
@@ -1,10 +1,16 @@
 {% extends "base.html" %}
 {% block title %}Edit {{ item.title }} — Shelf{% endblock %}
 {% block content %}
-<div class="max-w-2xl mx-auto">
+<div class="max-w-4xl mx-auto" data-item-edit-sections>
     <a href="/item/{{ item.id }}{% if back.key %}?from={{ back.key }}{% endif %}" class="text-shelf-muted hover:text-shelf-text text-sm mb-4 inline-block">&larr; Back to item</a>
 
-    <h1 class="text-2xl font-bold mb-6">Edit Item</h1>
+    <div class="flex items-center justify-between gap-4 mb-6">
+        <div>
+            <p class="text-xs uppercase tracking-wider text-shelf-muted">Edit item</p>
+            <h1 class="text-2xl font-bold mt-1">{{ item.title }}</h1>
+        </div>
+        <span class="text-xs uppercase tracking-wider text-shelf-muted">{{ media_types.get(item.media_type, item.media_type) }}</span>
+    </div>
 
     {# The save refused a value (#54). Nothing on the form was saved; the row's
        stored values render below. Copy is keyed on the funnel's error code —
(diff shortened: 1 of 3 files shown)


dgahagan:
Shipped in [v0.35.0](https://github.com/dgahagan/shelf/releases/tag/v0.35.0) — `docker pull dangahagan/shelf:0.35.0`. Thank you.

The detail that made this mergeable as-is was the `data-always-visible` escape hatch. Hiding fields by media type is the obvious half; the trap is that a row whose media type was corrected later, or was odd from the start, then has a real value sitting in a section the form refuses to show — invisible, uneditable, and still posted on save. Rendering that flag server-side per item, from whether the item actually holds a value, is the right place to decide it: the client cannot know, and a client-side guess would have been wrong in exactly the cases that matter. I would have asked for this if it had not been there.

Keeping every `name=` attribute and the form action untouched, and pinning that with a test that asserts the save contract is unchanged, is what let me ship a 304-line template rewrite in the same release as two other changes without a nervous re-read. That is the difference between a reorganisation and a rewrite.

**One consequence worth knowing about**, since it cost you elsewhere and was not your fault: this landed a few hours before the release and rewrote `static/js/item_edit.js`, which is the file your own #94 builds on — so #94 went `CONFLICTING` and I deferred it rather than merging it half-understood. I have commented there with what the rebase needs, plus a design question about the scanner overlay that is worth reading before you spend the effort.

**What I added:** a paragraph in `docs/user-guide/items.md` describing the six sections and the media-type behaviour, including the always-visible rule — the page still described Edit as one undifferentiated form. No code changes from me.

Issue #98, opened 2026-09-06
Add hierarchical physical locations


sudo-rpaisley:
## What problem would this solve?

Shelf's current location model is flat: an item can be in a location such as "Living Room", but real physical collections are often more precise.

A book might be in:

- Living Room → Bookcase 1 → Shelf 3
- Bedroom → Shelf 1
- Office
- Archive Room → Cabinet → Drawer

The hierarchy is not fixed. Some users only need a room, some need a shelf directly beneath a room, and others need several levels. Flat location names also make common labels such as "Shelf 1" ambiguous because the same name can legitimately exist under different parents.

This becomes particularly important with first-class physical copies (#97): location belongs to the individual physical copy, and two copies of one catalogue item may be stored in different places.

## What would you like to happen?

Allow physical locations to form an arbitrary parent/child tree.

A location node would have a name and an optional parent. This would allow structures such as room → bookcase → shelf without requiring those exact level names or a fixed depth.

Important behaviour:

- any level may be omitted;
- a location can be used directly even if it has children;
- the same child name may exist under different parents, e.g. Living Room → Shelf 1 and Bedroom → Shelf 1;
- duplicate sibling names under the same parent should be prevented case-insensitively;
- the UI should show the full path where ambiguity matters;
- existing flat Shelf locations should migrate safely as root nodes;
- physical copies should ultimately point to the precise location node;
- digital/service-backed availability should not receive a physical location.

I would also like the model to leave room for an optional physical ordering within a location. For example, titles on one shelf could later be ordered by drag-and-drop or automatically arranged by series/release order. That ordering does not need to be part of the first hierarchical-location PR, but the data model should not make it difficult later.

## Workarounds or alternatives you've considered

- Put the full path into one flat location name, e.g. "Living Room / Bookcase 1 / Shelf 3". This loses structure and makes browsing or moving a whole bookcase difficult.
- Create globally unique shelf names. This produces awkward names such as "Living Room Shelf 1" and "Bedroom Shelf 1" and still does not represent parent/child relationships.
- Add fixed Room, Bookcase and Shelf columns. This assumes every collection uses exactly three levels and does not handle cabinets, drawers, boxes or simpler layouts.

[...]

PR #99, opened 2026-09-06
Track first-class physical copies of catalogue items

### LINKED ISSUE #97: Track individual physical copies of an item

Shelf currently treats a catalogue item and the physical object on the shelf as essentially the same thing. That works while there is only one copy, but it breaks down as soon as someone owns two copies of the same title or edition.

Two physical copies can legitimately differ in condition, acquisition details, notes, barcode, provenance and location. They may also be lent separately. Storing those facts on the catalogue item means one copy overwrites the other, while creating duplicate catalogue items makes series, metadata and related-media views noisy.

This also makes it difficult to build the collector/inventory features already described on Shelf's roadmap, where condition, edition and provenance are properties of a particular owned object rather than of the abstract title.

Introduce first-class **physical copies** as child records of a catalogue item.

The catalogue item would continue to hold shared descriptive metadata such as title, creators, identifiers, synopsis, series and artwork. A physical copy would represent one owned object and could carry copy-specific facts such as:

- condition;
- acquisition date/source/price;
- copy-specific notes;
- inventory or local barcode;
- provenance;
- physical location;
- optional ordering/position within that location.

An item could have zero, one or many physical copies.

For ordinary users with one copy, the UI should stay simple: a single physical copy should not make routine browsing or editing feel more complicated. The extra structure becomes visible when the user needs to distinguish copies.

Digital availability should remain separate. A digital book, audiobook, comic or game hosted on a connected service is not a physical copy and should continue to be represented through the relevant digital/service relationship rather than being assigned a room or shelf.

This foundation should also make later features possible without baking them into the first change:

- lending a specific copy;
- hierarchical room → bookcase → shelf locations;
- physical shelf ordering;
- duplicate-copy inventory scans;
- per-copy valuation and provenance.

- Create duplicate catalogue items for each owned copy. This duplicates shared metadata and makes series/search/browse views noisy.
- Store one set of condition/location/acquisition fields directly on the item. This cannot represent two different copies.
- Encode copy details in free-text notes. This loses structure and cannot support inventory, lending or location workflows.

I have an implementation of copies/holdings in my fork, but I would prefer to agree the upstream data model and migration shape before preparing a PR because this becomes a foundation for several roadmap items.

sudo-rpaisley:
## What this changes

Adds a first-class physical-copy model beneath Shelf's existing catalogue items.

* An item remains the shared descriptive/catalogue record.
* `item_copies` represents individual physical objects.
* Each copy can hold its own condition, notes, acquisition date/source/price, provenance, location and local/accession barcode.
* Copy numbers are local to an item, while copy barcodes are globally unique.
* Items can have zero, one or many physical copies.
* Existing item-level location writes remain compatible by mirroring only the primary copy.
* Secondary copies are never silently moved by legacy item writes.
* Existing owned items are only backfilled when they already have an explicit physical location.
* No media type is guessed to be physical or digital from `owned` alone.

Fixes #97.

## Why

Shelf currently stores catalogue metadata and the ownership/location of a physical object on the same item row. That works for one copy, but cannot represent two copies of the same edition with different locations, condition, acquisition details or provenance.

Separating the catalogue item from its physical copies provides the foundation for duplicate copies, copy-specific lending, hierarchical locations, physical ordering and provenance without forcing digital or service-backed items into a physical-location model.

## Compatibility

The existing `items.location_id` field remains a compatibility projection during the transition. A primary copy mirrors that field; secondary copies remain independent.

The legacy backfill is deliberately conservative: only an owned item that already has an explicit location is treated as proven to have a physical copy. `owned=1` alone is not enough because digital and service-backed items may also be owned.

## Deliberately separate

This PR does not include copy-management UI, copy-specific lending, hierarchical locations (#98), physical shelf ordering, per-copy valuation, or digital/service holdings. Those can build on this foundation independently.

## Testing

The contribution branch has passed Shelf's full GitHub Actions validation, including unit/integration tests, offline lints, CSS reproducibility and the full browser E2E suite.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -147,6 +147,29 @@ MIGRATIONS: Sequence[tuple[int, str, str]] = (
                   AND isbn13 NOT GLOB '*[^0-9]*'
                   AND substr(isbn13, 1, 3) IN ('978', '979'))
         )"""),
+    (25, "Add first-class physical copies",
+     """CREATE TABLE IF NOT EXISTS item_copies (
+            id                 INTEGER PRIMARY KEY AUTOINCREMENT,
+            item_id            INTEGER NOT NULL REFERENCES items(id) ON DELETE CASCADE,
+            copy_number        INTEGER NOT NULL,
+            location_id        INTEGER REFERENCES locations(id) ON DELETE SET NULL,
+            condition          TEXT,
+            notes              TEXT,
+            acquired_date      TEXT,
+            acquisition_source TEXT,
+            acquisition_price  REAL CHECK(acquisition_price IS NULL OR acquisition_price >= 0),
+            provenance         TEXT,
+            copy_barcode       TEXT UNIQUE,
+            is_primary         INTEGER NOT NULL DEFAULT 0 CHECK(is_primary IN (0, 1)),
+            created_at         TEXT NOT NULL DEFAULT (datetime('now')),
+            updated_at         TEXT NOT NULL DEFAULT (datetime('now')),
+            UNIQUE(item_id, copy_number)
+        )"""),
+    (26, "Backfill primary copies from legacy locations",
+     """INSERT INTO item_copies (item_id, copy_number, location_id, is_primary)
+        SELECT i.id, 1, i.location_id, 1 FROM items i
+        WHERE i.owned = 1 AND i.location_id IS NOT NULL
+          AND NOT EXISTS (SELECT 1 FROM item_copies c WHERE c.item_id = i.id)"""),
 )
 
 MIGRATION_TABLES = """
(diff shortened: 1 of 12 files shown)


dgahagan:
Merged — this is a lot of careful work and the design judgement in it is the part I want to call out.

Specifically: making the backfill `owned = 1 AND location_id IS NOT NULL` rather than `owned = 1`, and writing down *why* in both the migration and the service docstring. `owned` alone would have quietly manufactured a physical copy for every service-backed digital row in the database, and that is the kind of migration you cannot take back once someone has upgraded. Same for refusing to infer physical-vs-digital from media type. And keeping `items.location_id` as a compatibility projection onto the primary copy means every existing add/edit/import/CSV surface keeps working without touching any of them — that seam is why this could land as one PR instead of ten.

The cycle detection in `_assert_parent_not_in_subtree` via a recursive CTE, and keeping `name` as the denormalised full path so browse and the select boxes need no changes, are both the right call.

**Two things for the record, neither of them blocking, both my problem to write up rather than yours to fix:**

1. **The description undersells the PR.** It says "This PR does not include ... hierarchical locations (#98)" — but it does: migrations 27–30, all 204 lines of `app/services/locations.py`, two test files, and `docs/user-guide/locations.md`. I merged it as both #97 and #98 and said so in the squash commit, so the history is accurate. I mention it only because I nearly reviewed it against the wrong scope, and the hierarchy work deserves the credit rather than arriving as a stowaway.

2. **`delete_location` is a user-visible behaviour change.** Deleting a location that has children now fails with `has_children` where it previously succeeded. That is the correct semantic — silently orphaning a subtree would be worse — but it is a change existing users will notice, so it needs a changelog line under Changed rather than riding along under Added. I'll write that.

Both #97 and #98 will close with the release. Thanks — and see my note on #94 about that bulk-move E2E failure, which turned out to be our bug, not yours.

dgahagan:
Shipped in **[v0.36.0](https://github.com/dgahagan/shelf/releases/tag/v0.36.0)** — this PR is most of the release. Thank you.

It is also the largest structural change Shelf has taken from a contributor: a new table, a new hierarchy on an existing one, five migrations and a backfill that writes rows. Some notes on what I checked and what I changed around it.

**The backfill decision is the right one, and I want to say so specifically.** Keying on `owned = 1 AND location_id IS NOT NULL` rather than `owned = 1` is what makes this migration safe, and it is not an obvious call from inside the code. I measured it on two databases before releasing: a real 1,057-item collection produced 1,006 primary copies, all correct; a dev seed with 1,078 owned items but only 66 located produced 66. Under the looser rule that second instance would have manufactured **1,012 physical copies for objects that do not exist**, in a table whose whole purpose is to describe objects that do. The comment in `docs/item-copies.md` calling `owned` insufficient evidence is doing real work.

**Verified live at both viewports** on a copy of the real collection, upgraded through the `_backfill_versions` path (a pre-version-tracking database, which replays every migration rather than taking the incremental route — the shape a long-standing user actually has). `schema_version` advanced to 30 with one log line and no traceback. The compatibility seam holds: changing an item's location through the ordinary edit form moved both `items.location_id` and its primary `item_copies` row, with no copy-aware code anywhere in that form. `ON DELETE RESTRICT` protects the data layer and not just the route — a direct `DELETE` of a parent raises `IntegrityError`.

**What I changed on the way out**, none of it in your code:

- `docs/item-copies.md` said hierarchical locations were "intentionally a separate follow-up (upstream issue #98)". True when you wrote it, false by the time the PR landed, since #98 shipped inside this same PR. Rewritten to describe the boundary that actually held.
- Both new doc pages were unlinked — neither was in `docs/README.md`'s area→page map, so nothing pointed at them. Added.
- `docs/architecture.md` is the authoritative structural description and had no `item_copies` at all, and described `locations` as a flat table. Both now documented, including why the denormalised path is worth a descendant rewrite.

**Two cosmetic things I did not fix and have filed instead**, so they are recorded rather than lost: the per-row **Inside** dropdown offers a node's own descendants (the server refuses correctly, but you have to submit to find out), and the create and edit forms label the same options differently — `Inside X` versus bare `X`.

One last thing worth knowing: CI on `main` was failing on `test_bulk_move_apply_moves_selected_list_item` when your PRs merged. That was not your doing — it is a pre-existing E2E race that the other half of this release fixes.

PR #100, opened 2026-09-06
Upstream/romm full

sudo-rpaisley:
## What this changes

Adds an optional RomM integration so Shelf can connect to a self-hosted RomM server, discover/synchronise games as Digital Game records, retain the stable RomM record identity, and surface the connected RomM item from Shelf.

The integration includes the RomM client, synchronisation and record-mapping layers, settings/UI support, pagination and retry handling, collision protection, sparse-response refresh safety, and restricted cover fetching.

## Why

RomM already provides the authoritative digital-game library for users who run it. This integration lets Shelf catalogue those games alongside the rest of a collection without duplicating the underlying ROM storage or treating a service-backed digital game as a physical holding.

## How it was tested

The automated provider-boundary tests use mocked RomM HTTP responses and cover the client, record mapping and full integration paths.

It has also been exercised against a live **RomM 5.2.0** instance containing **24,886 ROMs** across **65 configured platform entries / 63 distinct platform names**.

The library ranges from single-ROM platforms to several thousand titles on one platform, including:

* Sega Mega Drive/Genesis — 5,851 ROMs
* Wii — 1,759 ROMs
* Nintendo 3DS — 1,699 ROMs
* Game Boy — 1,585 ROMs
* Windows — 1,510 ROMs

This gives the platform-discovery/mapping, pagination and large-library synchronisation paths a substantial real-world workload in addition to the mocked automated coverage.

The current rebased head is running the contribution workflow now.

* [x] `make test` passes — current CI run in progress
* [x] `make test-e2e` passes — current CI run in progress
* [x] `make checks` passes — current CI run in progress
* [x] `make css` re-run, if templates or Tailwind classes changed — current CI run in progress

## Notes for review

Per the current review discussion, I have deliberately not respun the implementation while the shared RomM/Komga integration architecture is being settled — particularly outbound request pacing and the relationship between a service-synchronised item and Shelf's catalogue/holding model.

The earlier structural review points are noted for the eventual respin:

* explicit router registration in `main.py`;
* a real appended migration for `romm_records` rather than runtime schema creation;
* loading the item script from the item-detail template rather than through `components.js`.

I have left those implementation changes alone for now so the branch does not churn against assumptions that may change.
## What this changes

Adds an optional RomM integration so Shelf can connect to a self-hosted RomM server, discover/synchronise games as Digital Game records, retain the stable RomM record identity, and surface the connected RomM item from Shelf.

The integration includes the RomM client, synchronisation and record-mapping layers, settings/UI support, pagination and retry handling, collision protection, sparse-response refresh safety, and restricted cover fetching.

## Why

RomM already provides the authoritative digital-game library for users who run it. This integration lets Shelf catalogue those games alongside the rest of a collection without duplicating the underlying ROM storage or treating a service-backed digital game as a physical holding.

## How it was tested

The automated provider-boundary tests use mocked RomM HTTP responses and cover the client, record mapping and full integration paths.

It has also been exercised against a live **RomM 5.2.0** instance containing **24,886 ROMs** across **65 configured platform entries / 63 distinct platform names**.

The library ranges from single-ROM platforms to several thousand titles on one platform, including:

* Sega Mega Drive/Genesis — 5,851 ROMs
* Wii — 1,759 ROMs
* Nintendo 3DS — 1,699 ROMs
* Game Boy — 1,585 ROMs
* Windows — 1,510 ROMs

This gives the platform-discovery/mapping, pagination and large-library synchronisation paths a substantial real-world workload in addition to the mocked automated coverage.

The current rebased head is running the contribution workflow now.

* [ ] `make test` passes — current CI run in progress
* [ ] `make test-e2e` passes — current CI run in progress
* [ ] `make checks` passes — current CI run in progress
* [ ] `make css` re-run, if templates or Tailwind classes changed — current CI run in progress

## Notes for review

Per the current review discussion, I have deliberately not respun the implementation while the shared RomM/Komga integration architecture is being settled — particularly outbound request pacing and the relationship between a service-synchronised item and Shelf's catalogue/holding model.

The earlier structural review points are noted for the eventual respin:

* explicit router registration in `main.py`;
* a real appended migration for `romm_records` rather than runtime schema creation;
* loading the item script from the item-detail template rather than through `components.js`.

I have left those implementation changes alone for now so the branch does not churn against assumptions that may change.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -66,13 +66,16 @@ class SecurityHeadersMiddleware(BaseHTTPMiddleware):
             "default-src 'self'; "
             "script-src 'self'; "
             "style-src 'self' 'unsafe-inline'; "
-            "img-src 'self' data: blob: https:; "
-            "connect-src 'self'; "
+            "img-src 'self' data: blob: https:; "  # blob: for the intake photo preview
+            "connect-src 'self'; "  # PWA sync queue + SSE are same-origin only
             "font-src 'self' data:; "
             "base-uri 'self'; "
             "form-action 'self'; "
             "frame-ancestors 'none';"
         )
+        # Camera stays first-party for barcode scanning (html5-qrcode / ZXing
+        # getUserMedia) and the photo-intake webcam viewfinder; everything else
+        # is denied outright.
         response.headers["Permissions-Policy"] = (
             "camera=(self), microphone=(), geolocation=(), payment=(), usb=()"
         )
@@ -81,63 +84,102 @@ class SecurityHeadersMiddleware(BaseHTTPMiddleware):
 
 _SKIP_AUTH_PATHS = frozenset({"/login", "/setup", "/logout", "/health", "/sw.js"})
 _SKIP_AUTH_PREFIXES = ("/static/", "/covers/", "/share/")
+
+# Methods that mutate state and must carry a CSRF token
 _CSRF_METHODS = frozenset({"POST", "PUT", "PATCH", "DELETE"})
 
 
 class CSRFMiddleware(BaseHTTPMiddleware):
-    """Double-submit cookie CSRF protection for all state-mutating requests."""
+    """Double-submit cookie CSRF protection for all state-mutating requests.
+
+    HTMX sends the X-CSRF-Token header (configured in base.html).  Standard
+    browser form submissions are also blocked unless the hidden _csrf field or
+    header matches the csrf_token cookie.
+    """
 
     async def dispatch(self, request: Request, call_next) -> Response:
         path = request.url.path
+
+        # Only validate state-mutating methods on authenticated paths
         if request.method not in _CSRF_METHODS:
             return await call_next(request)
         if path in _SKIP_AUTH_PATHS or path.startswith(_SKIP_AUTH_PREFIXES):
             return await call_next(request)
+
         cookie_token = request.cookies.get("csrf_token")
         if not cookie_token:
             return Response("CSRF token missing", status_code=403)
+
+        # Accept token from header (HTMX) or form field (plain HTML forms).
+        # IMPORTANT: for form bodies we must cache the raw bytes BEFORE calling
+        # call_next, because BaseHTTPMiddleware's receive stream is consumed once.
+        # We replay the body via a patched Request so the route handler can still
+        # read the form data normally.
         submitted = request.headers.get("X-CSRF-Token")
         body_bytes: bytes | None = None
         if not submitted:
             content_type = request.headers.get("content-type", "")
             if "application/x-www-form-urlencoded" in content_type or "multipart/form-data" in content_type:
-                body_bytes = await request.body()
+                body_bytes = await request.body()  # caches in request._body
                 form = await request.form()
                 submitted = form.get("_csrf")
+
         if not submitted or submitted != cookie_token:
             return Response("CSRF validation failed", status_code=403)
+
+        # If we consumed the body, replay it so the route handler can read it too
         if body_bytes is not None:
             async def _replay_receive():
                 return {"type": "http.request", "body": body_bytes, "more_body": False}
+
             patched = Request(request.scope, _replay_receive)
             return await call_next(patched)
+
         return await call_next(request)
 
 
 class AuthMiddleware(BaseHTTPMiddleware):
     async def dispatch(self, request: Request, call_next) -> Response:
         from app.auth import get_current_user, should_refresh_token, set_auth_cookie, get_user_count
+
         path = request.url.path
+
+        # Skip auth for static assets
         if path.startswith(_SKIP_AUTH_PREFIXES):
             return await call_next(request)
+
+        # Inject user into request state
         user = get_current_user(request)
         request.state.user = user
+
+        # Setup wizard: if no users exist, redirect everything to /setup
         if path not in _SKIP_AUTH_PATHS:
             if get_user_count() == 0:
                 return RedirectResponse(url="/setup", status_code=303)
+
+        # Login redirect: if users exist but no session, redirect to /login
+        # (skip for POST /login, POST /setup to avoid blocking form submissions)
         if path not in _SKIP_AUTH_PATHS and not user:
             if get_user_count() > 0:
                 return RedirectResponse(url="/login", status_code=303)
+
         response = await call_next(request)
+
+        # Sliding expiry: refresh token if past half-life, preserving CSRF token
         if user:
             fresh_token = should_refresh_token(request)
             if fresh_token:
                 set_auth_cookie(response, fresh_token, request.cookies.get("csrf_token"))
+
         return response
 
 
 class RateLimitMiddleware(BaseHTTPMiddleware):
-    """Simple in-memory per-IP rate limiter for API endpoints."""
+    """Simple in-memory per-IP rate limiter for API endpoints.
+
+    Uses a bounded OrderedDict (max MAX_IPS entries) to prevent unbounded
+    memory growth from rotating IPs or IPv6 address churn.
+    """
 
     MAX_IPS = 1000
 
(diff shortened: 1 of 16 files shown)


dgahagan:
Holding this one for changes rather than merging, and the asks are structural rather than about the substance — the substance is good.

**What I like.** Keying persistence on the stable RomM id *only*, and explicitly refusing to adopt an existing Shelf game by title/platform similarity, is the right instinct — that is precisely how a cartridge quietly becomes a service-backed row and you headed it off. Raising when a RomM id is already attached to a non-RomM item rather than papering over it is the same good instinct. The paged `iter_rom_candidates` with retry on 408/429/5xx and a no-progress guard will matter on a large library. `_refresh_fields` not letting a later sparse response erase known metadata is a subtle one that most people miss.

And `_cover_target` deserves specific credit: resolving relative paths against the configured server, then rejecting any absolute URL whose hostname is not the configured RomM host, is exactly right. Provider metadata is untrusted input and you treated it that way without being asked.

**Three changes before I can take it:**

1. **Register the router in `main.py`, not `app/routers/__init__.py`.** `_pages.router.include_router(_romm.router)` from the package `__init__` makes registration a side effect of importing *any* `app.routers.*` module, and it nests `/api/romm` under the pages router rather than mounting it beside the other seventeen. Please add `app.include_router(romm.router)` to the block at `app/main.py:491-510` and leave `__init__.py` empty. It is one more line in core, which I know is what you were trying to avoid — but that explicit line is the thing that makes the mounting greppable.

2. **`romm_records` needs a real migration.** The module-level `_SCHEMA` run by `ensure_schema(db)` on every call is a table that no upgrade path, no `schema_version` row, and nothing reading `app/database.py` knows exists. I understand the reasoning in your comment — not consuming a migration number while other proposals are in flight — and it was a considerate thing to worry about, but the collision is mine to resolve, not yours to route around. #99 just merged and took 25–30, so please add `romm_records` as **31** in `MIGRATIONS` and add the same CREATE to `MIGRATION_TABLES`; then `ensure_schema` can go away entirely.

3. **Don't inject the item script from `components.js`.** The `document.createElement('script')` block keyed off a pathname regex runs on every page in the app to decide about one. Put `<script src="/static/js/romm-item.js"></script>` in the item detail template instead.

**Also worth knowing:** #101 does all three of the same things in the same three files, so it and this PR now conflict with each other on `app/routers/__init__.py`, `app/templates/settings.html` and `static/js/components.js`. I test-merged the queue and #100 goes in clean while #101 collides on exactly those. If you fix them here as above, most of that conflict evaporates on its own.

Rebase onto `main` when you pick this up — #94 and #99 both landed since you branched.

dgahagan:
Thanks for this, and for #101 alongside it — a RomM integration is something I want Shelf to have.

**Deferring this one past 0.36.0**, deliberately, and I want to say what it is waiting on rather than leave it sitting.

It is ~1,600 lines adding a whole new external service: a client, a sync engine, a record layer, a settings surface, three JS files and a docs page. That is bigger than anything Shelf has merged from a PR so far, and it lands in the part of the app I am most careful with — an outbound API client plus a sync that writes rows. Two specific things I need to settle before merging, neither of which is a criticism of the code:

1. **Outbound pacing.** Every metadata and cover client in Shelf goes through `app/services/outbound.py` (per-host minimum intervals, limits in `config.py`) rather than calling `httpx` directly. A self-hosted RomM is the user's own box, so it is arguable either way, and I would rather decide that once, for RomM and Komga together, than twice differently.
2. **How a synced item relates to a catalogued one.** Shelf already has one sync of this shape — Audiobookshelf — and 0.36.0 has just added first-class physical copies (your #99). I want the RomM and Komga models to agree with that, and with each other, before either ships. Reviewing them separately is how they end up disagreeing.

So: **I am reading #100 and #101 together as one design question**, not blocking on anything you need to change yet. Realistically that is the next week — I will come back here with the answer on those two points, and with anything the read turns up, before asking for changes.

One thing that would help in the meantime: the PR template came through unfilled. A couple of sentences on **what you ran it against** — which RomM version, how big a library — would tell me a lot that the diff cannot.

Sorry it is a wait. Two integrations of this size arriving on the same day is a good problem, and I would rather give them a proper read than rush one in.

dgahagan:

rewrite your work silently.

---

**On this PR specifically.** Same two things done right as #101 — the client token
is encrypted through `app/crypto.py`, and the mutating `fetch()` calls in
`romm-item.js` / `romm-settings.js` all carry `X-CSRF-Token`. And to answer a
question you didn't ask: **direct `httpx` use is correct here.** RomM is a
self-hosted server you're talking to, not a shared public metadata provider, so it
belongs with `audiobookshelf.py` rather than behind `outbound.py`'s pacing. No
change wanted there.

Beyond the shared ask, this one is the mildest of the six — `__init__.py` is just
an `include_router`, so moving it to `app/main.py` should be a two-line change.
The `romm_records` table is the other half.

**On the delay:** same as #101 — I deferred this at the 0.36.0 release without a
good enough explanation. The structural points above were the reason, and they're
the only thing between this and a merge.

dgahagan:
**Landed as [`c6541d1`](https://github.com/dgahagan/shelf/commit/c6541d1) (your authorship) plus a follow-up [`359a089`](https://github.com/dgahagan/shelf/commit/359a089).** Closing by hand rather than with the button because #113 landed first and this needed a conflict resolution I cannot push to your branch.

**On the table.** Your docstring says the integration table is local so it "does not consume an upstream migration number while other schema proposals are in flight". That hazard is real — you had five schema-carrying PRs open simultaneously, and I hit exactly the conflict you were avoiding while merging this batch. But `MIGRATION_TABLES` in `app/database.py` answers it without the tradeoff: it consumes no migration number either. It is a plain block of `CREATE TABLE IF NOT EXISTS` that `init_db` executescripts on every boot, fresh install and upgrade alike. Your DDL moved there verbatim; `ensure_schema` and its four call sites are gone.

Worth knowing for next time: a table created lazily on first request never appears in the upgrade path or the backup docs until someone exercises the feature, which is the part that bites later.

I have added `tests/test_schema_ownership.py` as a tripwire — no `CREATE TABLE` outside `app/database.py`, no `ensure_schema` helpers — so this is now enforced rather than left to review. Same for router registration in `tests/test_router_registration.py`.

The RomM code itself needed nothing. The no-adopt-by-title-similarity decision in `persist_candidate` is the right call and the docstring explaining it is why I did not have to guess.

dgahagan:

This sat deferred for a while, and the reason was never the code: #100 and #101 arrived as one design question — two self-hosted-library integrations with the same shape — and I wanted to answer it once rather than twice. Thank you for the patience.

**The refusal in this PR is the part I want to single out.** A RomM game is deliberately *not* matched onto an existing physical game by title and platform. That would silently turn someone's cartridge into a service-backed item, and the damage would be invisible until they went looking for it. Giving it its own record and letting the user link the two themselves is the correct answer, and it is the same instinct behind #101 keeping Komga's library kind apart from Shelf's media type. Both are now written into `docs/architecture.md` as stated refusals, so nobody "fixes" them later.

Keeping RomM's own record identity as the primary key of `romm_records` is what makes a re-sync update rather than duplicate — same reasoning, one layer down.

Two changes on the way in, both house rules you had no way to know:

- `romm_records` moved from a private `_SCHEMA` + `ensure_schema()` into `app/database.py`'s `MIGRATION_TABLES`. Creating tables outside that file bypasses `MIGRATIONS`/`schema_version`; it is **G1** in `GOTCHAS.md`. Four PRs in this batch did the same thing, so it was one decision, not a judgement about yours.
- Route registration moved out of `app/routers/__init__.py` and into `main.py`. Six PRs monkey-patched that file at import time and would have conflicted six ways with each other.

Credentials were correctly encrypted through `app/crypto.py` and every mutating `fetch()` sent `X-CSRF-Token` — noted, because those are the two things most likely to be missed and they were not.

RomM sync had no live server to test against on this network, so it is documented as verified by code and tests rather than driven end to end. If you run it against your own RomM and something is off, please do file it.

**On the merge:** the branch was reshaped too far for GitHub to build a merge commit, so it shows as closed. The work is in `main` and the public commits are authored to you — `c6541d15` and `359a0890`.

PR #101, opened 2026-09-06
Upstream/komga full

sudo-rpaisley:
## What this changes

Adds an optional Komga integration so Shelf can connect to a self-hosted Komga server, discover configured libraries, synchronise books/series into Shelf, retain the stable Komga record identity, and surface the connected Komga item from Shelf.

The integration includes Komga client/synchronisation and record-mapping layers, settings/UI support, library-level type classification, cautious library-kind suggestions, ISBN-aware matching, pagination handling, and refresh safety.

## Why

Komga already provides the authoritative digital comic/manga library for users who run it. This integration lets Shelf catalogue those service-backed books alongside the rest of a collection while keeping Komga as the source of the digital files.

## How it was tested

The automated provider-boundary tests use mocked Komga HTTP responses and cover book lookup, library classification, record mapping and the full integration paths.

It has also been exercised against a live **Komga v1.26.3-master** instance containing:

* **6,842 books**
* **365 series**
* approximately **245.4 GB** of library data

This gives library discovery, series/book synchronisation, library-level classification, pagination and large-library handling meaningful real-world coverage in addition to the mocked automated tests.

The current rebased head is running the contribution workflow now.

* [x] `make test` passes — current CI run in progress
* [x] `make test-e2e` passes — current CI run in progress
* [x] `make checks` passes — current CI run in progress
* [x] `make css` re-run, if templates or Tailwind classes changed — current CI run in progress

## Notes for review

Per the current review discussion, I have deliberately not respun the implementation while the shared RomM/Komga integration architecture is being settled — particularly outbound request pacing and the relationship between a service-synchronised item and Shelf's catalogue/holding model.

The earlier structural review points are noted for the eventual respin:

* explicit router registration in `main.py`;
* a real appended migration for `komga_records` rather than runtime schema creation;
* loading the item script from the item-detail template rather than through `components.js`;
* removing the fragile `app/routers/__init__.py` monkeypatching in favour of normal imports/central declarations.

Manga is intentionally not treated as a Komga-only product decision here. The separate first-class Manga media-type work is being handled independently in #102, so this integration can consume that type later without being the change that introduces it.

I have left those implementation changes alone for now so the branch does not churn against assumptions that may change.


dgahagan:
Holding this for changes too. It carries all three of the #100 asks — please read that comment first, since items 1–3 there apply here verbatim (router in `main.py`; `komga_records` as a real migration instead of `ensure_schema`; `<script>` tag in the item template instead of injection from `components.js`). This one has two more on top, both about it reaching outside its own integration.

**First, the good part.** Classifying at the *library* level rather than guessing per book is the right granularity, and `suggest_library_kind` suggesting Manga only on a clearly manga-named library while leaving the decision to the user is a good default. Matching on canonical ISBN rather than only your own id is a real improvement over the RomM equivalent.

**4. `app/routers/__init__.py` monkeypatches other modules' constants.** This:

```python
_series.UNASSIGNED_MEDIA_TYPES = tuple(_book_media_types)
_items_catalog.BOOK_MEDIA_TYPES = _book_media_types
```

works right now, and I checked why: nothing does `from app.routers.series import UNASSIGNED_MEDIA_TYPES`, so every reader picks up the patched value at call time. That is a coincidence, not a guarantee. The first person to add a from-import gets the pre-patch value and a bug where `manga` silently vanishes from one code path and not another — which is the exact trap the repo already documents for `items_common`. Please just edit the two literals in place at `app/routers/series.py:27` and `app/routers/items_catalog.py:185`, or repoint them at `BOOK_MEDIA_TYPES` with a normal import in those files. Happy to make that edit myself if you'd rather not respin twice for it — say the word.

**5. `manga` as a media type is a product decision, and it should not ride in on an integration.** Adding it to `MEDIA_TYPES` and `BOOK_MEDIA_TYPES` changes browse filters, scan detection, the CSV contract, valuation and every media-type dropdown, for every user, including everyone who will never run Komga. It may well be the right call — but it needs to be decided on its own merits and land in its own commit with its own changelog entry, so that a user reading the release notes sees "Shelf now has a Manga media type" rather than finding it inside "Komga integration".

Worth noting your own `komga_libraries.py` docstring says *"This first slice deliberately does not add new Shelf media types"* — so I think this may have grown past what you intended it to be.

I've opened #102 to decide the Manga question separately. If it lands, this PR gets simpler; if it doesn't, map Komga manga libraries onto `comic` and the integration still works. Either way I'd rather not have that decision gated on an integration review.

**Sequencing:** #100 first, then this one — they touch the same three files and #101 is the one that collides. Rebase onto `main` when you pick it up; #94 and #99 have landed since you branched.

Thanks for the batch. Four PRs in a day is a lot of work, and I'd rather ask for these changes now than have you discover them in a conflict later.

dgahagan:
Thanks for this — see my longer note on #100, because I am treating the two as one decision.

**Deferring past 0.36.0**, deliberately. Short version of the reasoning: #100 (RomM) and #101 (Komga) are ~1,600 and ~1,800 lines each, both adding a complete external-service integration — client, sync, records, settings surface, docs. Two questions apply to both, and I want to answer them once rather than twice differently:

1. **Outbound pacing** — whether a self-hosted service goes through `app/services/outbound.py` like the metadata clients do, or calls out directly like the Audiobookshelf sync does.
2. **How a synced item relates to a catalogued one** — Shelf's one existing sync of this shape is Audiobookshelf, and 0.36.0 has just landed first-class physical copies (your #99). RomM, Komga and ABS should agree on that model.

This one has two things #100 does not, which I will flag now so nothing is a surprise later:

- It touches `app/config.py` and `app/services/detect.py` — shared files rather than new ones. `detect.py` in particular feeds the scan path, which is the highest-traffic code in the app, so that hunk gets read closely.
- It is the larger of the two and carries `-10` lines of existing code.

Same timeline as #100: I am reading them together, realistically over the next week, and I will come back here with the answer on those two points before asking for changes. Nothing for you to do yet.

If you have a moment, the PR template came through unfilled — which Komga version and roughly how big a library you tested against would be genuinely useful.

Sorry for the wait, and thanks for the volume of work you have put into Shelf lately.

dgahagan:

rewrite your work silently.

---

**On this PR specifically.** Two things done right that I want to acknowledge,
since they're the parts most contributions get wrong: the API key goes through
`app/crypto.py` (`encrypt_value` / `decrypt_value` / `get_encryption_key`) and is
genuinely encrypted at rest, and every mutating `fetch()` in `komga-item.js` and
`komga-settings.js` sends `X-CSRF-Token`. Testing against a real 6,842-book
instance is also the reason I trust the pagination and matching code.

Beyond the shared ask:

- The `__init__.py` here reassigns globals in two *other* routers
  (`series.UNASSIGNED_MEDIA_TYPES`, `items_catalog.BOOK_MEDIA_TYPES`). Those two
  literals should genuinely be centralised — that's what **#105** does — so the
  cleanest route is: settle #105, let it land the config change, and drop these
  two lines from here entirely.
- This depends on the Manga media type from #105, which is held on my decision in
  **#102**. I'll answer that issue first.

**On the delay:** I deferred this and #100 at the 0.36.0 release and I owe you a
clearer answer than I gave then. The blocker was never the feature — it's the two
structural points above, and they're the same two for both PRs. Once they're
settled these are near the front of the queue.

dgahagan:
**Landed as [`969248f`](https://github.com/dgahagan/shelf/commit/969248f) (your authorship) plus a follow-up [`3bbf253`](https://github.com/dgahagan/shelf/commit/3bbf253).** Closed by hand rather than with the button — #100 and #113 landed first, so this needed conflict resolution I cannot push to your branch. The conflicts were all pure unions: `app/main.py`, `item_detail.html` and `settings.html` each take both RomM's and Komga's line.

**The Manga projection worked exactly as you designed it, and that is worth saying.** You wrote `_shelf_media_type` as a conditional:

```python
if kind == "manga" and "manga" in MEDIA_TYPES:
    return "manga"
return "comic"
```

#105 landed earlier in this same batch and added `manga` to `MEDIA_TYPES`. Your integration immediately started projecting a manga library onto `manga` instead of falling back to `comic` — no code change needed. Keeping the provider `kind` separate from Shelf's media type, which is what your "KOMGA: keep library kind separate from Shelf media type" commit did, is the reason this was three stale descriptions rather than a redesign.

What I updated: two tests still asserting `media_type == "comic"`, and the `_reclassify_owned_record` docstring still saying Shelf maps both kinds to `comic`. The logic is untouched.

**On the table.** Same as #100 — `komga_records` and its two indexes moved into `MIGRATION_TABLES` in `app/database.py`. That consumes no migration number, so the concern in your RomM docstring about not burning one while other schema PRs are in flight is fully met, without the table being invisible to the upgrade path. `tests/test_schema_ownership.py` now enforces it.

dgahagan:

Read as one design question alongside #100, which is why both sat deferred for a while. Thank you for waiting it out.

**Keeping the library's Comic/Manga kind apart from Shelf's own media type is the decision that makes this safe.** `komga_records.kind` is checked in the schema and is not `items.media_type`, so flipping a library's kind on the Komga side cannot reclassify something the user catalogued by hand. That is now written into `docs/architecture.md` as a stated refusal so it does not get "simplified" away later. Matching on ISBN where one exists, and only there, is the same restraint.

This PR also had a dependency chain behind it: it needed Manga to exist as a Shelf media type, which needed #105, which needed a decision in #102 that had not been made. #102 was decided yes on 2026-09-07 — rationale is shelving, not taxonomy — and that unblocked both.

Three changes on the way in:

- `komga_records` moved from a private `_SCHEMA` + `ensure_schema()` into `MIGRATION_TABLES` in `app/database.py` (**G1** — creating tables outside that file bypasses `MIGRATIONS`/`schema_version`). Four PRs in the batch needed the same rework.
- Route registration moved from `app/routers/__init__.py` into `main.py`; the PR reassigned `series.UNASSIGNED_MEDIA_TYPES` and `items_catalog.BOOK_MEDIA_TYPES` at import time, which six PRs did in six mutually incompatible ways.
- The Manga projection was settled against the shipped `BOOK_MEDIA_TYPES` rather than the reassignment.

Credentials encrypt correctly through `app/crypto.py`, and every mutating `fetch()` carries `X-CSRF-Token`.

Like RomM, Komga sync had no live server available on this network, so it ships verified by code and tests rather than driven end to end. Please file anything you hit against your own server.

**On the merge:** the branch was reshaped too far for GitHub to build a merge commit, so it shows as closed. The work is in `main`, authored to you — `969248ff` and `3bbf253f`.

Issue #102, opened 2026-09-07
Decide whether Shelf should have a Manga media type


dgahagan:
Split out of #101 (Komga integration), which adds `manga` to `MEDIA_TYPES` and `BOOK_MEDIA_TYPES` as part of the integration.

The addition may well be correct, but it is a product decision with reach well beyond Komga: a new media type changes the browse filters and their cross-filter counts, scan detection's book-family hints, the CSV import/export contract, valuation grouping, and every media-type dropdown — for every user, including everyone who will never run a Komga server.

That should be decided on its own merits and land in its own commit with its own changelog entry, so the release notes say "Shelf now has a Manga media type" rather than burying it inside an integration.

**The question:** is manga a distinct media type, or a comic with different provenance?

Arguments for a distinct type:
- Different reading direction, different collecting and shelving conventions, different metadata providers.
- Users who collect both will want to filter them apart.
- Komga itself distinguishes manga libraries, so the signal exists upstream.

Arguments against:
- `comic` already covers "sequential art with an ISBN and a series", and every book-family code path treats them identically.
- It is one more type in every dropdown for a distinction many users do not make.
- Splitting later is a migration; merging later is also a migration. Neither is free.

If we do add it, it needs: the `MEDIA_TYPES` and `BOOK_MEDIA_TYPES` entries, the two local book-family literals in `app/routers/series.py:27` and `app/routers/items_catalog.py:185` repointed at the config constant, an icon, and a docs pass.

If we don't, #101 should map Komga's manga libraries onto `comic` — the integration works either way.

dgahagan:
Decided: **yes, adding it.**

The reasoning is shelving, not taxonomy. Manga is a distinct enough thing to
browse and filter on that people who own any will want it separated, and the
`MEDIA_TYPES` list exists to answer "how do I find this on my shelf", not to
classify a medium. So this is not a ruling on the relationship between manga and
comics, and I am deliberately not making one — anyone who would rather file theirs
as Comic can keep doing exactly that.

It goes in the book family (`BOOK_MEDIA_TYPES`) on the same practical grounds as
the rest of that family: it is read, it carries an ISBN, and it belongs to a
series.

Shipping via **#105** from @sudo-rpaisley.

dgahagan:
Decided yes, and shipped in [v0.37.0](https://github.com/dgahagan/shelf/releases/tag/v0.37.0) via #105 from @sudo-rpaisley.

[...]

Issue #31, opened 2026-08-21
Unassigned series


LegendaryB:
### What problem would this solve?

Show books which have no series assigned.

### What would you like to happen?

A "dummy" series entry which holds all books which have no series assigned. Shown under series as Unassigned

dgahagan:
Shipped in [v0.13.0](https://github.com/dgahagan/shelf/releases/tag/v0.13.0) — thanks for the report, @LegendaryB.

Books with no series now appear on the Series page in an **Unassigned** block after your real series. The heading carries the true total ("1014 books with no series"), and the strip below shows a sample of twelve covers with "· showing 12" in the count line when there are more. Click any cover to open the item and set a series from its edit page.

One design decision that went a different way than a straight "treat it as a series" would suggest: the block is deliberately *not* a series. It has no rename, disband, mark-complete, synopsis or Hardcover-check controls, it is excluded from the `Series (N)` heading count, it never appears in the rename autocomplete, and it sorts last rather than by size — an unassigned pile is usually the biggest group in a library, and it shouldn't become the headline of a page about series. For the same reason the Complete and Incomplete filter chips hide it: a pile of unsorted books makes no claim about completeness either way.

Scope is books (`book`, `kids_book`, `audiobook`, `ebook`, `comic`) — CDs, DVDs and video games essentially never carry a series, so including them would bury the books you were looking for.

`docker pull dangahagan/shelf:0.13.0`

PR #105, opened 2026-09-07
Add Manga as a first-class media type

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test

sudo-rpaisley:
## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -13,13 +13,14 @@ MEDIA_TYPES = {
     "dvd": "DVD / Blu-ray",
     "cd": "CD",
     "comic": "Comic / Graphic Novel",
+    "manga": "Manga",
     "video_game": "Video Game",
 }
 
 # The book family: media types that are read, carry ISBNs, and belong to a
 # series. Everything else in MEDIA_TYPES (dvd, cd, video_game) is a disc or a
 # cartridge. Declared here, beside the types themselves.
-BOOK_MEDIA_TYPES = frozenset({"book", "kids_book", "audiobook", "ebook", "comic"})
+BOOK_MEDIA_TYPES = frozenset({"book", "kids_book", "audiobook", "ebook", "comic", "manga"})
 
 # Seed data — runtime platform list comes from game_platforms table
 GAME_PLATFORMS = {
--- a/app/routers/items_catalog.py
+++ b/app/routers/items_catalog.py
@@ -14,7 +14,7 @@ from fastapi import APIRouter, Depends, Form, Request
 from fastapi.responses import HTMLResponse, JSONResponse
 
 from app.auth import require_role
-from app.config import HTTP_TIMEOUT, MEDIA_TYPES
+from app.config import BOOK_MEDIA_TYPES, HTTP_TIMEOUT, MEDIA_TYPES
 from app.database import get_db, get_game_platforms, get_setting
 from app.routers import items_common
 from app.services import covers, igdb, openlibrary, scan_outcome, tmdb
@@ -181,8 +181,6 @@ async def add_game_from_search(
     )
     resp.headers["HX-Trigger"] = items_common._toast_header(f"Added: {metadata['title'][:50]}")
     return resp
-
-BOOK_MEDIA_TYPES = {"book", "kids_book", "audiobook", "ebook", "comic"}
 
 @router.get("/title-search")
 async def title_search(
--- a/app/routers/series.py
+++ b/app/routers/series.py
@@ -9,6 +9,7 @@ import logging
 from fastapi import APIRouter, Depends, Form, Request
 
 from app.auth import require_role
+from app.config import BOOK_MEDIA_TYPES
 from app.database import gc_orphaned_series_meta, get_db, get_setting
 from app.services import hardcover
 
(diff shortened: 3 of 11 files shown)


dgahagan:
**#102 is decided — yes, manga gets its own media type — so this is going in.**
Reasoning is on the issue; short version is that it is a shelving decision rather
than a taxonomic one, and I am not adjudicating manga-vs-comic.

I checked how far the change actually reaches and it holds up well: every
media-type dropdown iterates `media_types.items()` (Browse, Scan, Intake, Edit,
`filter_counts_oob.html`) and `intake.BOOK_SEARCH_MEDIA_TYPES` aliases
`cover_queue.COVER_REQUEUE_MEDIA_TYPES`, so Browse, Scan, Intake and the intake
book-search path all follow from the `MEDIA_TYPES` entry plus your `cover_queue`
change. Nothing else to wire.

Three things before merge:

**1. One place the change doesn't reach.** `item_detail.html:111` gates the
*Fetch synopsis* button on an inline `('book', 'ebook', 'audiobook', 'kids_book')`
literal, and this PR doesn't touch that file — so manga would inherit **comic's**
behaviour and get no synopsis button. That may well be right (Open Library and
Google Books synopsis coverage for manga volumes is patchy in the same way it is
for comics), but I'd rather it were a decision than an omission. What did you see
in your own testing — is a synopsis fetch worth offering for manga?

**2. The `series.py` decision you reversed — you had no way to know.** That file
said, verbatim: *"Deliberately a local literal, not an import of
items.BOOK_MEDIA_TYPES / synopsis.BOOK_MEDIA_TYPES (which already disagree about
comics) — see docs: issue #31 design plan §1."* The plan it cites isn't public,
which is my gap, not yours.

Having re-derived it: **your version is right, and this decision is why.** With
manga in the picture, three hand-maintained copies of the book family is exactly
how manga silently drops out of series assignment or scan detection later. Keeping
them in sync by hand was already the weaker option; adding a sixth member settles
it. The comment goes, the import stays.

**3. But the collision it warned about is still live, and wants a rename rather
than a merge.** After this PR there are still two constants called
`BOOK_MEDIA_TYPES`: yours in `config.py` (frozenset, includes `comic` and now
`manga`) and `synopsis.py`'s (4-tuple, no `comic`), which `cover_queue.py` imports
and `item_detail.html:111` hand-copies a third time.

Those two aren't a duplication to collapse — they answer different questions.
*"Is this in the book family?"* and *"will we fetch a synopsis for it?"* have
genuinely different answers for comics, which is what the old comment was getting
at. The actual defect is that they share a name. I'd take
`synopsis.BOOK_MEDIA_TYPES` → `SYNOPSIS_MEDIA_TYPES`, with `item_detail.html`
reading it instead of restating it — which also fixes (1) properly, whichever way
that question goes.

Happy for that rename to be a follow-up PR rather than growing this one; your
call.

**Small thing:** `cover_queue.py`'s docstring loses its indentation in this diff —
`after a full pass with nothing ready we sleep to the earliest deadline.` ends up
at column 0. Looks accidental.

One sequencing note, no action needed from you: `manga` here, `magazine` in #106
and the four music types in #109 all land in `config.py`'s `MEDIA_TYPES`. Manga is
settled; the other two aren't yet, so expect me to ask you to rebase this one or
those, depending on what merges first.

dgahagan:
**Landed as [`99938c5`](https://github.com/dgahagan/shelf/commit/99938c5) with your authorship** — closing this as merged rather than through the button, because #112 landed first and this needed a conflict resolution I could not push to your branch.

**Where I took a different route.** #112 split the Identifiers section into two Alpine components — `x-data="isbnCamera"` for ISBN-carrying types and `x-data="upcCamera"` for retail barcodes — so the outer `#edit-identifiers` section no longer owns the camera. Your diff added `manga` to five `data-media-types` lists; I added it to **seven**, including the inner `isbnCamera` div at line 214 that did not exist when you cut this branch. Without that one, Manga would have rendered the Identifiers section with no ISBN field in it.

I deliberately left `manga` out of the retail list (`dvd cd comic video_game`) — manga carries ISBNs, not UPCs, which matches how you scoped `BOOK_MEDIA_TYPES`.

Two other small things I fixed rather than send back: the branch stripped the trailing newline from `item_edit.html` and `detect.py`.

Nothing here was wrong on your side — #112 simply moved underneath you.

dgahagan:
Shipped in [v0.37.0](https://github.com/dgahagan/shelf/releases/tag/v0.37.0).

This one was blocked on a decision rather than on code — #102, "should Shelf have a Manga media type at all". Decided yes on 2026-09-07, and the rationale is worth stating because it is narrower than the question sounds: **shelving, not taxonomy.** Shelf does not need to adjudicate what counts as a comic; it needs to let someone keep a manga run apart from a run of graphic novels, which is a thing people actually do with physical shelves. Putting Manga in the book family so it carries an ISBN, belongs to a series and filters separately is exactly the right expression of that.

Small but real: this unblocked #101, because Komga libraries carry their own Comic/Manga kind and there was nothing on the Shelf side to map the manga half onto.

One thing changed on the way in — the PR reversed a recorded decision in `series.py` where `UNASSIGNED_MEDIA_TYPES` is "deliberately a local literal … see issue #31 design plan §1". That comment is the only place the reasoning lives, so there was no way to know from the outside. The merged version keeps the literal and adds `manga` to `BOOK_MEDIA_TYPES` in `config.py` instead, which is where the family sets now live.

`docs/configuration.md`'s "media types are a fixed list" was stale after this and has been brought up to date — it was missing magazine, vinyl, cassette and digital music as well.

PR #106, opened 2026-09-07
Rebase full Periodicals contribution onto Shelf 0.36.0

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test

sudo-rpaisley:
## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.


dgahagan:

rewrite your work silently.

---

**On this PR specifically.** The barcode reasoning is sound and I want to keep it:
a 977 carrier identifies the publication, not an issue, so a carrier-only scan
must not match an existing issue or the next issue becomes a false duplicate.
That's a real insight and it should survive the rework. Adding `magazine` to
`MEDIA_TYPES` and pacing the ISSN portal through `outbound.fetch` are both right.

This is the PR where the import-time patching goes furthest — `__init__.py`
replaces three upstream functions (`items_common._scan_upc`,
`items._find_item_by_barcode`, `periodicals.find_periodical_item`). Each of those
is a real behavioural change to the scan path, and each deserves to be an explicit
edit at its own call site where a reviewer will find it. I'd rather work through
the right seam for those three with you than have you guess — happy to do that on
this thread.

Note also that `magazine` lands in the same `MEDIA_TYPES` decision described on
#109; I'll settle that before any of the three merges.

dgahagan:
Follow-up on the media-type question I left open above: **`magazine` is approved,
and so is the model behind it.**

To be specific, because it is the larger half: I am not just taking the
`MEDIA_TYPES` entry and deferring the rest. Periodicals as **publication +
concrete issue records** is approved as the design, including the two tables. I
considered taking `magazine` as a plain catalogue row and leaving the model for
later, and rejected it — a magazine you can file but cannot track issues against
is a weak feature, and it would have meant asking you to tear out the part of this
PR that is actually the interesting work.

The 977-carrier reasoning is the reason I am comfortable with it. A carrier
identifies the publication rather than an issue, so a carrier-only scan must not
resolve to an existing issue or the next issue becomes a false duplicate. That is
a real distinction that a simpler model would have gotten wrong, and it should
survive the rework intact.

So the only things between this PR and a merge are the two structural points from
my earlier comment. One adjustment to those: since the periodicals model is
approved, `periodical_publications` and `periodical_issues` should go into the
`MIGRATIONS` tuple as part of that same rework rather than as a separate step —
no need to treat the schema as provisional.

**Please still hold on rebasing.** #109's four music types are in the same
`config.py` `MEDIA_TYPES` block and I have not settled those yet, so a rebase now
risks being wasted. I will come back to you on #109 shortly; `manga` (#105) and
`magazine` are both settled, so it is only the music formats outstanding.

dgahagan:
**The rebase hold is lifted** — you can rebase this whenever you are ready.

I settled the last media-type question on #109 about twenty minutes after my
comment here, and I answered it on that thread rather than this one, so this PR
was still showing a hold that no longer applied. Apologies for the wasted wait.

Where the `MEDIA_TYPES` block landed, in full, so you do not have to reconstruct
it from three threads:

- `manga` — approved (#105)
- `magazine` — approved, along with the publication + issue model behind it (my
  comment above)
- `vinyl`, `cassette`, `digital_music` — approved (#109); **`music_other` dropped**

That is the whole of it. Nothing further is pending on `config.py`.

One thing to expect rather than act on: `docs/configuration.md` states the media
types verbatim ("Media types are a fixed list: book … video game") and
`docs/user-guide/browse-and-search.md` has the same list in its Type table. All
three of #105, #106 and #109 touch that block, so whichever of them lands second
and third owes a small docs rebase. Not a reason to hold anything.

Still outstanding on this PR, unchanged from my earlier comment: router
registration in `app/main.py`, and `periodical_publications` /
`periodical_issues` through the `MIGRATIONS` tuple. The three upstream functions
replaced at import time (`items_common._scan_upc`, `items._find_item_by_barcode`,
`periodicals.find_periodical_item`) are the part I would rather work through with
you than have you guess at — happy to do that here whenever you want to start.

dgahagan:
**Landed as [`45a6b4d`](https://github.com/dgahagan/shelf/commit/45a6b4d) (your authorship) plus a follow-up [`02c6743`](https://github.com/dgahagan/shelf/commit/02c6743).** Closed by hand — #105 and #113 landed first, so this needed conflict resolution I cannot push to your branch.

**Where I did not take your side of a conflict.** In `app/config.py` you removed `manga` from `BOOK_MEDIA_TYPES` and generalised the comment above it. The removal is only because you branched before #105 (Manga) landed earlier in this batch, so I kept `manga`. On the comment you were right that it needed to change — `magazine` makes the old parenthetical `(dvd, cd, video_game)` wrong — but your replacement dropped information, so the merged version names the discs and cartridges *and* points at `PERIODICAL_MEDIA_TYPES`:

```python
# The book family: media types that are read, carry ISBNs, and belong to a
# series. Everything else is deliberately outside this family: discs and
# cartridges (dvd, cd, video_game), and periodicals, which are issue-based
# and carry their own family below. Declared here, beside the types themselves.
```

Naming `PERIODICAL_MEDIA_TYPES` as a separate family rather than stretching `BOOK_MEDIA_TYPES` over magazines was the right structural call, and the docstring explaining why you named the family rather than the format — room for journals and newspapers later — is what made it obvious.

**On the tables.** `periodical_publications` and `periodical_issues` and their three indexes moved into `MIGRATION_TABLES`; `ensure_schema` is gone along with four call sites in the service and three in `app/routers/periodicals.py`. See my note on #100 for why `MIGRATION_TABLES` meets the migration-number concern from your RomM docstring.

One test could not survive the move: `test_periodical_schema_is_idempotent` was testing `ensure_schema` itself. Rather than drop it I rewrote it as `test_periodical_tables_come_from_migration_tables`, which asserts the same property — both tables present, replaying the DDL harmless — so the coverage you wrote is still there.

dgahagan:

**Separating the publication from the issue is the whole feature**, and it is the thing a naive implementation gets wrong: a run of one magazine as one publication with many issues, rather than forty unrelated rows that happen to share a title. Naming the family separately from the `magazine` format — so journals and newspapers can join later without rewriting every consumer — is thinking a version ahead, and `PERIODICAL_MEDIA_TYPES` shipped exactly as you shaped it.

**The refusal here is the best decision in the PR.** A 977 barcode's supplement digits are not reliable enough to fill the issue number and date unattended, and you resolve the publication from the ISSN and then stop, leaving those two fields for the user to confirm. That is the honest answer and it is now recorded in `docs/architecture.md` so it does not get "improved" into a guess.

Two changes on the way in, both house rules with no external signal:

- `periodical_publications` and `periodical_issues` moved from a private `_SCHEMA` + `ensure_schema()` into `MIGRATION_TABLES` in `app/database.py` (**G1**: creating tables there bypasses `MIGRATIONS`/`schema_version`). Four PRs in the batch needed the same rework.
- Route registration moved into `main.py`. This PR's `routers/__init__.py` replaced `items_common._scan_upc`, `items._find_item_by_barcode` and `periodicals.find_periodical_item` at import time; six PRs rewrote that same empty file in six incompatible ways.

Pacing through `outbound.fetch` was already correct — noted, because it is easy to reach for `httpx` directly.

One caveat on verification: the ISSN portal returned a blanket 403 to every ISSN tried from this network, which looks like a datacenter-IP block on their side rather than anything in the code. `issn_portal.lookup()` handled it as a clean `found=False` with no crash, which is the right degradation, but it means the happy path is unverified against the live service here. If it behaves differently from your network, I would like to know.

**On the merge:** the branch was reshaped too far for GitHub to build a merge commit, so it shows as closed. The work is in `main`, authored to you — `45a6b4d4` and `02c6743e`.

dgahagan:
Follow-up: **Periodicals shipped in 0.37.0 with no navigation entry**, so the page rendered correctly at its URL and nothing in Shelf linked to it. For #108 that meant the Home overview the release notes described was not reachable at all — login and the brand link both went to Browse.

PR #107, opened 2026-09-07
Rebase Settings layout onto Shelf 0.36.0

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test

sudo-rpaisley:
## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.


dgahagan:
Credit first: this is the only PR in the batch that re-ran `make css` and
restamped `static/sw.js`'s `SW_VERSION` (`vb04af0d1` → `vac51ee2a`) without being
asked. That's the generated-artifact rule followed exactly, and it saves me a
step.

**One bug to fix.** The person icon in the new `fragments/account_menu.html` has a
truncated path — the body of the figure is missing:

```
this PR:   d="M16 7a4 4 0 11-8 0 4 4 0 018 0zM12 14a7 7 0 00-7-7z"
base.html: d="M16 7a4 4 0 11-8 0 4 4 0 018 0zM12 14a7 7 0 00-7 7h14a7 7 0 00-7-7z"
```

The `7h14a7 7 0 00` segment is gone, so the avatar glyph renders broken. Looks
like it got clipped while moving the markup rather than an intentional change.

**One question.** Settings is now gated on `user.role == 'admin'` inside the
account menu. I want to confirm that matches what non-admin users saw before
rather than tightening access as a side effect of a layout change — can you say
what the behaviour was in your testing for an editor and a viewer account?

The move itself I like — the nav row was overflowing and Settings is the right tab
to pull out of it.

dgahagan:
**Landed as [`e5049ae`](https://github.com/dgahagan/shelf/commit/e5049ae) with your authorship** — closed by hand, as the last of eleven PRs merged today.

**The conflict.** #100 and #101 added `romm.html` and `komga.html` beside `integrations.html` in `settings.html`, while this PR wrapped the includes in `<div data-testid="settings-section-content">`. I kept your wrapper and put all six fragments inside it, with romm and komga next to integrations where those PRs placed them.

**On the generated assets.** `static/css/app.css` and `static/sw.js` conflicted with #115, which is unavoidable — you both edited build output. They are regenerated with `make css` rather than merged, and `SW_VERSION` is restamped to `vdc269f36` from the new precache digest. If you hit this again on a future PR, the fix is always to re-run `make css` on the merged tree; neither side of such a conflict is ever the right answer.

Worth flagging one thing I nearly got wrong myself: I restored a trailing newline on `app.css` *after* stamping, which silently invalidated the digest and failed `TestSwPrecacheDigest`. Generated files have to be the last thing touched.

Also kept: the admin-only pin on Settings access from your "SETTINGS: pin existing admin-only Settings access" commit, and the restored account icon path.

dgahagan:

Settings had grown into one long page and splitting it into Library / Integrations / Data / Users is the change that makes it navigable again — particularly this release, which added RomM and Komga forms to it. Remembering the section you were last in, in the browser rather than on the account, is the right storage choice for a preference like that. And being explicit in the PR that this is navigation and presentation only, with every form, endpoint and stored value behaving as before, is what made it reviewable at all next to twelve other PRs.

Two things fixed before it landed:

- **A corrupted person-icon SVG** — the path read `M12 14a7 7 0 00-7-7z`, with the `7h14a7 7 0 00` body segment missing. Almost certainly an editor or copy-paste artifact rather than anything you wrote.
- Settings was gated on `role == 'admin'` directly. Shelf keeps that check in one place; Settings remains administrator-only, and account and password controls stay under **Account**, which is where a non-admin needs them.

This is one of the two PRs in the batch I flagged as needing you rather than needing a fix — thank you for turning the rebase onto 0.36.0 around as fast as you did.

**On the merge:** the branch was reshaped too far for GitHub to build a merge commit, so it shows as closed. The work is in `main`, authored to you — `e5049ae9`.

PR #108, opened 2026-09-07
Rebase Home layout onto Shelf 0.36.0

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test

sudo-rpaisley:
## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.


dgahagan:

rewrite your work silently.

---

**On this PR specifically.**

- **The red CI is not your fault.** It fails on
  `test_browse_sort_restored_in_new_session`, which is exactly the flake your own
  **#110** fixes — that test asserted the default `newest` order differs from
  `title_desc`, but `newest` has no tie-break for rows inserted in the same
  second. I'm merging #110 first; rebase on it and this should go green.
- The `__init__.py` here **deletes the root `GET /` route out of
  `pages.router.routes[:]`**. Replacing the root redirect with a real home page is
  a change I want — but as an edit to the route in `pages.py`, not as a filter
  applied to another module's route list at import.

dgahagan:

Home answers a question Browse was never shaped to answer — "what is happening in my library?" versus "find me this thing" — and keeping those two jobs on separate pages is the right split. Totals, what is lent out, missing covers, the media-type breakdown and recent additions is a well-chosen set: every one of them is a prompt to do something.

**Building it from existing state on each open, storing nothing new, is what makes it cheap to own.** No cache to invalidate, no counter to drift out of step with reality. The Scan item shortcut for editors and admins is a nice touch.

This PR arrived red, on exactly the Browse sort-restore flake that #110 fixes — not on anything in it. #110 merged first for that reason, and this rebased onto it cleanly.

Route registration moved from `app/routers/__init__.py` into `main.py`. This one's version deleted the root `GET /` out of `pages.router.routes[:]` at import time to take over the index; six PRs rewrote that same empty file in six incompatible ways, so all six moved to the `main.py` convention #112 was already using.

dgahagan:
Follow-up: **Home shipped in 0.37.0 with no navigation entry**, so the page rendered correctly at its URL and nothing in Shelf linked to it. For #108 that meant the Home overview the release notes described was not reachable at all — login and the brand link both went to Browse.

Issue #81, opened 2026-09-02
music catalogue: MusicBrainz + Cover Art Archive, with physical formats


sudo-rpaisley:
# music catalogue: MusicBrainz + Cover Art Archive, with physical formats

I have been working on  music catalogue support in my fork and would like to contribute the work upstream if the direction fits Shelf.

This directly addresses the gap noted in #44, whose description identifies roadmap item **6b — MusicBrainz + Cover Art Archive** as the long-term solution for music/CD metadata.

## Proposed core

The implementation treats the exact owned release or pressing as the Shelf item and adds:

* Vinyl, Cassette, CD, Digital Music and Other Music as first-class media types
* a relational music model beneath `items` for releases, media/discs/sides, tracks and pressing identifiers
* MusicBrainz **Release** search by title, artist, barcode and catalogue number
* Cover Art Archive release-specific artwork
* Release Group linking through the existing `item_links` mechanism for alternate owned formats
* music-aware detail/edit UI with square artwork
* track listings preserving positions such as A1/B1
* conservative Vinyl/Cassette detection alongside existing CD detection
* enrichment of an already scanned title-only music item to an exact MusicBrainz Release without replacing the Shelf item

MusicBrainz remains the canonical identity source. `items` remains Shelf's generic owned object, so existing locations, lending, ownership/wishlist and related behaviour continue to work normally.

## Collector metadata

I also have a separate optional Discogs enrichment layer working for exact physical pressings, including Release/Master IDs, matrix/runout and other pressing identifiers, genres/styles, notes, token configuration and attribution.

I would propose keeping Discogs as a separate follow-up PR so the MusicBrainz/Cover Art Archive core does not depend on it.

## Contribution approach

The complete implementation is running and passing the full test suite in my fork.

Because my fork has diverged substantially from upstream, I would not submit that branch directly. If this design is acceptable, I will prepare a clean branch from current `dgahagan/shelf:main`, transplant only the music functionality, add the required `[Unreleased]` changelog entry, run the upstream-required test/check targets, and submit it using the project PR template.

Would you prefer the initial contribution to include Vinyl, Cassette, CD and Digital Music together, or start with CD + MusicBrainz/Cover Art Archive and add the other formats incrementally?

dgahagan:
Shipped in [v0.37.0](https://github.com/dgahagan/shelf/releases/tag/v0.37.0) — and shipped by you, in #109. Thank you.

[...]

PR #109, opened 2026-09-07
Add first-class Music catalogue on Shelf 0.36.0

### LINKED ISSUE #123: Release gate depends on a live UPC Item DB trial quota: split the decision test from the contract test

sudo-rpaisley:
## How it was tested

- [ ] `make test` passes
- [ ] `make test-e2e` passes
- [ ] `make checks` passes

## Notes for review

---

  drop input.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -106,20 +106,56 @@ MAX_TILES_PER_REQUEST = 16
 # Image input tokens ~= (w * h) / 750, capped per image at the model max.
 IMAGE_TOKEN_DIVISOR = 750
 ANTHROPIC_HIGHRES_IMAGE_TOKEN_CAP = 4784
+ANTHROPIC_STANDARD_IMAGE_TOKEN_CAP = 1600
 
-# --- External API rate limits ---------------------------------------------
-# Minimum seconds between requests to the same host. Used by
-# app/services/outbound.py. Values are deliberately a hair conservative.
-HOST_RATE_LIMITS = {
-    "openlibrary.org": 0.11,
-    "covers.openlibrary.org": 0.11,
-    "www.googleapis.com": 0.11,
-    "api2.isbndb.com": 1.05,
-    "hardcover.app": 0.12,
-    "api.hardcover.app": 0.12,
-    "images-na.ssl-images-amazon.com": 0.5,
-    "api.igdb.com": 0.25,
-    "api.themoviedb.org": 0.1,
+# USD per million tokens (input, output), matched by substring on model id.
+# Output dominates cost and scales with book count, not tile count.
+VISION_PRICING = {
+    "fable-5": (10.00, 50.00),
+    "opus": (5.00, 25.00),
+    "sonnet": (3.00, 15.00),
+    "haiku": (1.00, 5.00),
+}
+VISION_PRICING_DEFAULT = (5.00, 25.00)
+PROMPT_OVERHEAD_TOKENS = 500  # unified spine+cover prompt + JSON schema scaffolding
+TOKENS_PER_BOOK = 60  # ~1 JSON row, e.g. {"title": "Dune", "authors": "Frank Herbert", "isbn": "9780441172719", "source": "read"}
+EXPECTED_BOOKS_PER_MEGAPIXEL = 8  # rough spine density for output estimate
+EXPECTED_BOOKS_MIN = 20
+EXPECTED_BOOKS_MAX = 200
+
+# Minimum seconds between requests to a given outbound host, enforced by
+# app.services.outbound.acquire(). A host that is absent means "no pacing"
+# (interval 0.0) — that is the correct entry for LAN/self-hosted targets and
+# for hosts that publish no limit, not an oversight.
+#
+# Read this table as `app.config.HOST_RATE_LIMITS` at call time. A
+# `from app.config import HOST_RATE_LIMITS` freezes the reference at import
+# and breaks test overrides (the "Config import trap" in CLAUDE.md); tests
+# override single hosts with monkeypatch.setitem.
+HOST_RATE_LIMITS: dict[str, float] = {
+    # Open Library grants 1 req/s by default and 3 req/s to requests that
+    # identify themselves with an app name AND contact information
+    # (https://openlibrary.org/developers/api). openlibrary.py sends a
+    # contact URL in its User-Agent, so 3/s applies. If that header ever
+    # loses its contact info, this must drop to 1.0.
+    "openlibrary.org": 0.34,
+    # Two services behind one host: CoverID/OLID-keyed URLs are unlimited,
+    # every other key (ISBN, LCCN, OCLC) is capped at 100 requests per IP
+    # per 5 minutes and returns 403 past it
+    # (https://openlibrary.org/dev/docs/api/covers). 403 is not transient,
+    # so exceeding this reads as "no cover found" and fails silently — a
+    # per-host interval cannot tell the two key types apart, so pace for the
+    # limited one. Do not lower below 3.0.
+    "covers.openlibrary.org": 3.0,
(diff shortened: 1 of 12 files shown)


dgahagan:

rewrite your work silently.

---

**On this PR specifically.** The outbound work is exactly right and I want to say
so: `musicbrainz.org: 1.05` registered in `HOST_RATE_LIMITS`, a real identifying
`User-Agent`, and every provider call routed through `outbound.fetch`. That's
MusicBrainz's published policy met properly, and it's the first thing I checked.

This one also carries a product decision I need to make before it can land, so
don't rebase it until I come back to you:

- **Four new media types** — `vinyl`, `cassette`, `digital_music`, `music_other`.
  That's the largest single change to `MEDIA_TYPES` anyone has proposed, and it
  collides in `config.py` with `manga` (#105) and `magazine` (#106) from this same
  batch. I need to decide the shape of the media-type list across all three
  before any of them lands, rather than deciding it three times.
- `docs/configuration.md` says "Media types are a fixed list: book … video game"
  and this PR doesn't update it, so merging as-is would ship wrong docs. Worth
  fixing whichever way the decision goes.
- Four new tables, per the migrations note above — this is the largest of the four
  on that front.

This answers your **#81**, and the scoping question I left open there is really
this media-type question. I'll answer it on the issue so the discussion is in one
place.

dgahagan:
Coming back on the media-type question, which is the last one outstanding —
`manga` (#105) and `magazine` (#106) are both settled now.

**Approved: `vinyl`, `cassette` and `digital_music`. Please drop `music_other`.**
That makes a music family of four with the existing `cd`.

Since I spent a while on this and landed somewhere different from where I
started, the reasoning is worth having:

**The argument against your version is one you have already answered in the
code.** `music_media` stores a row per medium with the verbatim MusicBrainz
`format` string, and `music_releases.format_summary` keeps the release-level one —
so the format is modelled relationally, at a fidelity `items.media_type` cannot
carry. `_infer_media_type()` returns on the first recognised format, which means
an LP with a bonus 7", or a CD+DVD edition, gets a single label chosen by
iteration order. On a clean slate I would put one `music` type in `MEDIA_TYPES`
and a `format` field on `items` alongside `platform`, with a `BrowseFilter` for it.

**What decided it the other way is `cd`.** It is already a top-level media type
with user data behind it, so any single-`music`-type design either leaves `cd` as
an inconsistent sibling or needs a migration rewriting live rows in an
append-only schema. The tidier model would still owe a `cd` exception at the end
of it, and it is not worth rebuilding a working 1,347-line PR to get there. Your
`MUSIC_MEDIA_TYPES` including `cd` is the right call given where Shelf already is,
and the comment you left saying so is accurate.

The shelving argument also holds better here than I first credited: vinyl and CDs
live in different furniture, which is the same test that got `manga` and
`magazine` their rows. That is not true of hardcover vs paperback, which is why
Shelf has never had those.

**On `music_other`.** "Other Music Format" is a weak row in a dropdown every
user sees, including people who own no music at all, and it exists only as
`_infer_media_type`'s fallback for MiniDisc, SACD, DAT and so on. Two ways to
replace it and I would rather you picked, since you have the live library:

- fall back to `cd` — wrong sometimes, but correctable in one click and honest
  that the type is a shelving hint rather than a claim; or
- leave the type unset and let the user choose on the add screen.

**A question that could still change my answer**, so worth asking before you
rework anything: in your own collection, how many releases have more than one
medium *format* — an LP plus a 7", a CD plus a DVD? If that is common rather than
rare, the lossy projection stops being cosmetic and the `format`-field design gets
a lot more attractive. I would rather hear the real number than guess.

**Two other things for the rebase:**

- `docs/configuration.md` lists the media types verbatim ("Media types are a fixed
  list: book … video game") and this PR does not update it, so merging as-is would
  ship wrong docs. `docs/user-guide/browse-and-search.md`'s Type table is the
  same. #105 is touching both for `manga`, so whichever of you lands second owes
  the rebase — no action needed yet, just so it is not a surprise.
- **The hold is lifted.** All three media-type questions are settled, so you can
  rebase this whenever you are ready. It is still behind the registration and
  migrations rework from my earlier comment — four tables here, the largest of the
  batch on that front.

dgahagan:
**Landed as [`4fa8200`](https://github.com/dgahagan/shelf/commit/4fa8200) (your authorship) plus a follow-up [`3b50490`](https://github.com/dgahagan/shelf/commit/3b50490).** Closed by hand — #105, #106 and #113 landed first, so this needed conflict resolution I cannot push to your branch.

**The schema move is the one your own docstring asked for.** `music_catalog.ensure_schema` said:

> This first upstream slice owns its schema locally so it can be validated without consuming an upstream migration number before the maintainer has agreed the final music model. A production integration can move these statements into Shelf's normal bootstrap once the shape is accepted.

The shape is accepted, so I did that move: four tables and eight indexes into `MIGRATION_TABLES`. Worth knowing for next time — `MIGRATION_TABLES` never consumes a migration number, so it was always available for this without the "validate locally first" step. It is a plain `CREATE TABLE IF NOT EXISTS` block that `init_db` runs on every boot.

**Two things I fixed that were genuinely outstanding.** The CI failure here was two assertions in `test_music_full_integration.py` still expecting `music_other` after you removed it — you pinned the new behaviour in `test_music_foundation.py` but the older file was missed.

And `app/routers/__init__.py` still composed the feature at package import, including `music` into `pages.router` and mutating the cover allow-list via `music_covers.register_cover_art_archive()`. Your other PRs in this batch — #100, #101, #106, #108 — had all moved off that pattern; this one had not. It is now registered in `app/main.py` like the rest, `music_covers.py` is deleted, and `coverartarchive.org` is declared in `covers.ALLOWED_COVER_DOMAINS` beside the other twelve. Paths are unchanged, since both routers are prefix-free.

**What I did not change.** `_infer_media_type` returning `None` for an unrecognised format, so the add screen asks rather than guessing CD, is the right call and the docstring saying so is why I left it alone. Same for reusing the existing `cd` type rather than minting a second CD identity — the comment on `MUSIC_MEDIA_TYPES` made that decision legible.

dgahagan:
Shipped in [v0.37.0](https://github.com/dgahagan/shelf/releases/tag/v0.37.0), and it closes the issue you filed for it, #81 — the fuller write-up is over there.

Short version: this is the largest single feature in the largest release Shelf has had, and it is yours. Cataloguing the *release* rather than the title is the decision the whole thing rests on — two pressings of one album keeping their own country, date, label, catalogue number, packaging and medium, with track lists that preserve real side and track numbering so a vinyl A1/B1 stays A1/B1. Linking releases that share a MusicBrainz release group so the vinyl and the CD of one album find each other is the detail that makes it feel finished. Asking the user to choose when MusicBrainz reports an unrecognised medium, rather than guessing, is the right instinct throughout the batch.

Cover Art Archive came with it, which strikes an internal roadmap item that had been parked since August waiting for someone to actually want music support.

Two changes on the way in, both explained at length in #81: the four tables moved into `app/database.py`'s `MIGRATION_TABLES` (**G1**), and the fourth new media type folded into Shelf's existing `cd` rather than creating a second CD identity that every pre-existing CD row would have been on the wrong side of.

Pacing was already right — `musicbrainz.org: 1.05` registered with a real User-Agent, on their published policy. That is the kind of thing that usually has to be asked for.

**On the merge:** the branch was reshaped too far for GitHub to build a merge commit, so it shows as closed. The work is in `main`, authored to you — `4fa82002` and `3b504906`.

dgahagan:
Follow-up: **Music shipped in 0.37.0 with no navigation entry**, so the page rendered correctly at its URL and nothing in Shelf linked to it. For #108 that meant the Home overview the release notes described was not reachable at all — login and the brand link both went to Browse.

PR #111, opened 2026-09-07
EDITING: expose manual cover URL in Artwork

sudo-rpaisley:
## What this changes

Makes Shelf's existing secure manual-cover URL action available where users expect it: **Edit → Artwork**.

- adds a Cover image URL field and explicit **Use URL** action to the existing Artwork section;
- reuses Shelf's existing SSRF-safe `manual_cover` downloader rather than introducing a second download path;
- applies the selected cover immediately without submitting the main item-edit form;
- suppresses the normal item-detail `HX-Redirect` when the action originates from Edit, so unrelated unsaved edits remain in the DOM;
- leaves the existing item-detail cover-picker behaviour unchanged;
- keeps the URL input out of the normal item-write whitelist and explicitly includes only that value in the HTMX action, avoiding nested forms or a second metadata persistence path.

## Why

Shelf already supported downloading a cover from a URL, but the control was not exposed from the sectioned item editor. Users editing artwork therefore had file upload available in the obvious place while the existing manual-URL capability was effectively hidden elsewhere.

## How it was tested

Focused regressions cover the Edit markup and redirect-preservation behaviour. The current upstream CI run is green.

- [x] `make test` passes
- [x] `make test-e2e` passes
- [x] `make checks` passes
- [x] `make css` re-run

## Notes for review

This is deliberately a small UI exposure of an existing hardened backend. It does not add a new cover-fetch implementation or change ordinary item-edit persistence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Shelf are documented here. The format follows
 
 ## [Unreleased]
 
+### Added
+
+- **Manual cover URLs are now available directly in Edit → Artwork.** Paste a
+  public HTTPS image URL and choose **Use URL** to apply it immediately through
+  Shelf's existing SSRF-safe manual-cover downloader. The action stays on the
+  edit page, so unsaved metadata is not discarded; normal file uploads and
+  other item edits still wait for **Save Changes**.
+
 ## [0.36.0] - 2026-09-07
 
 Shelf's locations were one flat list. *Office*, *Bookcase 1* and *Shelf 3* sat
--- a/app/routers/items_covers.py
+++ b/app/routers/items_covers.py
@@ -230,6 +230,7 @@ async def cover_select(
 async def cover_from_url(
     item_id: int,
     url: str = Form(...),
+    return_to: str | None = Form(None),
     _=Depends(require_role("editor")),
 ):
     """Use a user-pasted public HTTPS image as an item's cover."""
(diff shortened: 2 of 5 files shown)


dgahagan:
Merging. This is the model for how to write one of these: a filled-in description
that says what it does and why, a CHANGELOG entry, and it reuses the existing
SSRF-safe `manual_cover.download` instead of adding a second download path.

I checked the claim in your description that the new `<input name="url">` is safe
inside `#item-edit-form`, since a stray field in that form is the obvious risk —
it holds. `update_item` reads a fixed key whitelist (`items.py:1185`), so Save
Changes ignores the field entirely. Good instinct to call it out; that's exactly
the kind of note that makes a review fast.

The `return_to=edit` flag to suppress `HX-Redirect` is the right shape too — the
item-detail cover picker keeps its existing behaviour and nothing forks.

This conflicts with #105 and #112 on `item_edit.html`. That's my problem to
sequence, not yours; no action needed.

dgahagan:

Manual cover URLs existed since 0.35.0 but only from the cover picker; putting **Use URL** directly in Edit → Artwork is what makes them reachable at the moment you actually want them. Keeping the action on the edit page so unsaved metadata is not discarded was the right call and is why it needed no rework.

One thing checked before merging, since it looked riskier than it was: the stray `url` input reaching `update_item` is ignored, because that function works from a fixed key whitelist. No path for it to be written to a column.

The changelog entry for this originally shipped without your name on it — my error, fixed before the tag.

Issue #90, opened 2026-09-04
Bare 12-digit legacy book UPC silently files a DVD


dgahagan:
## What happens

Scanning a legacy Scholastic price-point book barcode **without** its five-digit
supplement — just the 12-digit UPC-A, e.g. `078073003501` — silently files a
DVD. On a real collection it filed:

> **"Clifford s First Snow Day" — DVD / Blu-ray via upc**

That is a genuine product-database hit for that UPC, not a fallback, which is
exactly why it is so convincing and so wrong. The book in hand was a Baby-Sitters
Club paperback.

## Why it happens

`legacy_book.parse()` requires 17 digits, so a 12-digit read never enters the
legacy path added in #88. It falls through to `upc_svc.detect_barcode_type()` →
`"upc"` → the ordinary product lookup, which is unchanged and behaves as it
always has.

This is **pre-existing**, not a regression from #88 or #82. I checked whether
#82's check-digit fix newly enabled it: it did not — `validate_upc` had no
caller in `app/` before #88 became its first consumer.

## Why it matters

Many 1D scanners need explicit configuration to read a five-digit add-on. A user
whose scanner does not read it gets a confidently wrong DVD row with no
indication that a better answer existed, and #88's whole benefit is invisible to
them.

Shelf is not helpless here: `078073` is in `_PUBLISHER_PREFIXES`, so at the
moment it files the DVD it already knows it is looking at a legacy *book*
barcode.

## What it cannot be

The supplement **is** the title number. Without it there are on the order of
100,000 candidates behind that UPC, not two — so no lookup can resolve a bare
legacy UPC to the right book. The only available improvement is to stop filing a
confidently wrong one.

## Constraint — do not simply reject the scan

`CONTRIBUTING.md:63-66`: an unreadable barcode **deliberately creates a
placeholder row instead of being rejected**, because the offline scanner drops a
flushed code from its queue and rejecting it would silently lose a scan someone
made in a shop. Refusing to file would trade a wrong row for a lost scan, which
is worse.

## Suggested shape

When a bare 12-digit UPC's prefix is in `_PUBLISHER_PREFIXES`, still create the
row, but file it as an unresolved **book** carrying a card along the lines of
"older book barcode — rescan including the five digits to its right, or scan the
printed ISBN", instead of running the product lookup that returns a DVD.

**Open question for design:** whether the offline Store-mode queue path shares
this code, which decides whether this is one branch or two. Worth checking
rather than assuming.

## Reproduction

[...]

PR #112, opened 2026-09-07
EDITING: scan and edit retail UPC/EAN barcodes

sudo-rpaisley:
## What this changes

Extends the existing item-editor identifier controls so retail UPC/EAN barcodes can be entered or scanned directly.

- adds UPC/EAN controls to the existing **Identifiers** section;
- uses Shelf's existing camera/scanner engine;
- scanning fills the field only — it does not auto-save or trigger metadata lookup;
- canonicalises UPC-A to EAN-13 for storage;
- validates checksums;
- rejects 978/979 ISBNs from the retail-barcode field;
- retains Shelf's existing per-media duplicate protection;
- leaves ordinary item-edit persistence authoritative for the rest of the form.

## Why

Shelf already has strong ISBN editing/scanning behaviour, but physical media such as games, music, films and other retail products also need their normal UPC/EAN identifier to be editable without re-running an Add workflow.

## How it was tested

Focused tests cover validation, canonicalisation, duplicate handling and the non-submitting camera behaviour. Upstream CI is green.

- [x] `make test` passes
- [x] `make test-e2e` passes
- [x] `make checks` passes
- [x] `make css` re-run

## Notes for review

Magazine carrier/add-on handling is deliberately left to the Periodicals work rather than coupling ordinary retail barcode editing to that feature.


dgahagan:
This is the best-built PR in the batch and I want to be specific about why, since
it's the pattern I'm asking for elsewhere: it **registers its router in
`app/main.py`**, the way the project documents. It also validates both checksums,
canonicalises UPC-A to EAN-13 so storage identity is single-valued, refuses
978/979 Bookland carriers from the retail field, reuses the existing scanner
engine rather than adding a second, and closes the check-then-write duplicate race
with an `IntegrityError` fallback. That last one is a race most PRs don't notice.

I also checked whether nesting `data-media-types` inside the Identifiers section
breaks the show/hide — it doesn't. `updateEditSectionVisibility` uses a flat
`querySelectorAll` and sets `.hidden` per element, so the nested blocks evaluate
independently. Fine as written.

Two fixes before merge:

- **The refusal paths bypass the form's error surface.** The invalid-barcode and
  duplicate branches return bare `HTMLResponse("Invalid UPC / EAN barcode", 400)`
  and `HTMLResponse("Update conflicts…", 409)`. Every other refusal in
  `update_item` goes through `_refused(code)`, which redirects back to the edit
  form with `?error=<code>` so the template renders a banner and the user keeps
  their form — that's G58, and the comment explaining it sits about eight lines
  above your new code. As written the user gets a bare error string and loses what
  they'd typed. Two new error codes and the template copy should cover it.
- **`app/routers/item_barcode_edit.py` is only reachable from its own tests.**
  Nothing in the UI calls `GET /api/items/{id}/barcode-context`, and
  `_canonical_upc` is documented as a "compatibility wrapper used by focused
  barcode-edit tests". I'd rather not add a public API route to give a test
  something to import — can the tests call `upc.canonical_retail_barcode`
  directly, and the route come back when something actually uses it?

Nit: triple blank line in `upc.py` above `validate_ean13`.

This is also relevant to **#90** (a bare 12-digit legacy book UPC silently files a
DVD) — not the same fix, but it's the same identifier seam, so I'll look at them
together.

dgahagan:

Worth saying explicitly: **this PR was the one that showed the batch the right convention.** Six other PRs in it registered their routes by rewriting `app/routers/__init__.py` with import-time monkey-patching; #112 registered in `main.py` like the rest of the codebase does. That made the fix for the other six a known destination rather than an argument, and it is the reason that rework was mechanical.

Two changes on the way in:

- The two refusal paths returned bare `HTMLResponse` 400/409 instead of going through `_refused(code)`. That is **G58** in `GOTCHAS.md`, documented eight lines above the code in question — an easy thing to miss reading the surrounding function rather than the file.
- `/api/items/{id}/barcode-context` was dropped. Its only consumer was its own test, and an endpoint with no caller is a maintenance liability rather than a feature.

Everything else — UPC-A canonicalised to EAN-13, checksum validation, refusing 978/979 Bookland codes because they belong in the ISBN field, and scanning that fills the field without saving or triggering a lookup — shipped as you wrote it. Refusing Bookland in particular is the kind of decision that prevents a whole category of confused bug report.

PR #113, opened 2026-09-07
SHELF FILL: rapid placement into hierarchical locations

sudo-rpaisley:
## What this changes

Adds a dedicated **Shelf Fill** workflow for rapidly filing physical media into a precise hierarchical Shelf location.

- keeps the selected room/bookcase/shelf sticky while scanning multiple items;
- supports camera scanning through Shelf's existing scanner engine;
- moves existing catalogue items without repeating metadata-provider lookups;
- delegates unknown ISBN/UPC barcodes to Shelf's normal Add pipeline;
- supports copy-specific barcodes so duplicate copies can be placed independently;
- promotes wishlisted physical items to owned when they are shelved;
- keeps the primary copy and compatibility `items.location_id` field in sync through Shelf's normal write funnel;
- rejects inappropriate digital-only placement;
- remains compatible with plain Shelf 0.36;
- when copy `position_order` support is present, scans naturally append in physical scan order.

## Why

Hierarchical locations make it possible to describe a precise physical place, but filling a real shelf one item at a time through the normal editor is unnecessarily slow. Shelf Fill turns that into a continuous scanning workflow while reusing the existing catalogue and metadata logic.

## How it was tested

Focused tests cover hierarchical placement, wishlist promotion, exact duplicate-copy placement, scanner behaviour and compatibility paths. The complete upstream CI run is green.

- [x] `make test` passes
- [x] `make test-e2e` passes
- [x] `make checks` passes
- [x] `make css` re-run

## Notes for review

This 0.36 implementation deliberately uses the native `locations` and `item_copies` models rather than carrying forward the fork's older location/holdings compatibility layer.

### DIFF
--- a/app/routers/__init__.py
+++ b/app/routers/__init__.py
@@ -1,23 +1,10 @@
 """Router package composition for the Shelf Fill contribution."""
 
-from app import nav as _nav
 from app.routers import pages as _pages
 from app.routers import shelf_fill as _shelf_fill
 
+# Keep Shelf Fill self-contained. The page is intentionally not injected into
+# the global navigation registry here: upstream pins that registry exactly, and
+# the feature remains reachable directly at /shelf-fill without mutating shared
+# presentation state during package import.
 _pages.router.include_router(_shelf_fill.router)
-
-if not any(tab.get("key") == "shelf_fill" for tab in _nav.NAV_TABS):
-    insert_at = next(
-        (idx + 1 for idx, tab in enumerate(_nav.NAV_TABS) if tab.get("key") == "scan"),
-        len(_nav.NAV_TABS),
-    )
-    _nav.NAV_TABS.insert(insert_at, {
-        "key": "shelf_fill",
-        "label": "Shelf Fill",
-        "path": "/shelf-fill",
-        "roles": ("admin", "editor"),
-    })
-    _nav.HIDEABLE_TABS = [
-        tab for tab in _nav.NAV_TABS if tab["key"] not in _nav.ALWAYS_VISIBLE
-    ]
-    _nav.HIDEABLE_KEYS = frozenset(tab["key"] for tab in _nav.HIDEABLE_TABS)
--- a/app/routers/__init__.py
+++ b/app/routers/__init__.py
@@ -1,0 +1,23 @@
+"""Router package composition for the Shelf Fill contribution."""
+
+from app import nav as _nav
+from app.routers import pages as _pages
+from app.routers import shelf_fill as _shelf_fill
+
+_pages.router.include_router(_shelf_fill.router)
+
+if not any(tab.get("key") == "shelf_fill" for tab in _nav.NAV_TABS):
+    insert_at = next(
+        (idx + 1 for idx, tab in enumerate(_nav.NAV_TABS) if tab.get("key") == "scan"),
+        len(_nav.NAV_TABS),
+    )
+    _nav.NAV_TABS.insert(insert_at, {
+        "key": "shelf_fill",
+        "label": "Shelf Fill",
+        "path": "/shelf-fill",
+        "roles": ("admin", "editor"),
+    })
+    _nav.HIDEABLE_TABS = [
+        tab for tab in _nav.NAV_TABS if tab["key"] not in _nav.ALWAYS_VISIBLE
+    ]
+    _nav.HIDEABLE_KEYS = frozenset(tab["key"] for tab in _nav.HIDEABLE_TABS)
--- a/app/routers/shelf_fill.py
+++ b/app/routers/shelf_fill.py
@@ -31,6 +31,35 @@ def _copy_by_barcode(db, raw: str):
     return dict(row) if row else None
 
 
+def _has_position_order(db) -> bool:
+    """True when the optional physical shelf-ordering follow-up is installed."""
+    return any(
+        row["name"] == "position_order"
+        for row in db.execute("PRAGMA table_info(item_copies)").fetchall()
+    )
+
+
+def _append_copy_position(db, copy_id: int | None, location_id: int) -> int | None:
+    """Append one copy to the selected shelf when ordering support exists.
+
+    Shelf Fill remains independently usable on plain 0.36.0: older schemas do
+    not have ``position_order`` and simply skip this compatibility hook.
+    """
+    if copy_id is None or not _has_position_order(db):
+        return None
+    next_position = db.execute(
+        "SELECT COALESCE(MAX(position_order), 0) + 1 AS n FROM item_copies "
+        "WHERE location_id = ? AND id != ?",
+        (location_id, copy_id),
+    ).fetchone()["n"]
+    db.execute(
+        "UPDATE item_copies SET position_order = ?, updated_at = datetime('now') "
+        "WHERE id = ? AND location_id = ?",
+        (next_position, copy_id, location_id),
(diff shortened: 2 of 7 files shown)


dgahagan:

rewrite your work silently.

---

**On this PR specifically.** The feature is good and the reasoning in the
description is exactly right — reusing the existing catalogue and Add pipeline
instead of a parallel one, writing through `update_item_fields` rather than around
it, promoting wishlist → owned on shelving. That's the design I'd have asked for.

Two things beyond the registration change:

- **The page has no way in.** You note this deliberately ("intentionally not
  injected into the global navigation registry … reachable directly at
  `/shelf-fill`"), and I understand not wanting to touch shared presentation
  state. But a feature users can only reach by typing the URL isn't really
  shipped. I'd like a nav entry — happy for that to be my change if you'd prefer
  to leave the registry alone.
- **`_has_position_order()` runs `PRAGMA table_info(item_copies)` on every scan**
  to detect a `position_order` column that doesn't exist in this repo. I'd rather
  not pay a schema probe per request for a follow-up that isn't here yet — can
  that come out, and come back with the PR that adds the column?

dgahagan:
**I took the registration rework on this one myself** — you did not answer either
way on my offer, and I would rather you spent your time on #106 and #109, where
the rework is real work, than on this one, where it is two lines. Nothing is
being asked of you on this PR.

Here is the whole conversion, so you can see it is not hiding anything.
`app/routers/__init__.py` goes back to empty, and `app/main.py` gets:

```diff
-from app.routers import pages, items, items_covers, items_csv, items_catalog, locations, platforms, settings, ...
+from app.routers import pages, items, items_covers, items_csv, items_catalog, locations, shelf_fill, platforms, settings, ...

 app.include_router(locations.router)
+app.include_router(shelf_fill.router)
 app.include_router(platforms.router)
```

That is it. This is what I meant by "#112 shows it's a small change" — for a PR
that only adds routes, the documented path costs two lines and produces no
`__init__.py` for the next contribution to collide with.

**On the nav entry — you were right, and I want to say so specifically.** Your
comment said you left the registry alone because "upstream pins that registry
exactly." I went to add the tab and `tests/test_nav.py` has
`test_registry_covers_the_nine_tabs`, which asserts the key list exactly. So the
registry really is pinned, you read it correctly, and declining to mutate it
during package import was the right instinct rather than an oversight.

The answer is that the pin is a guard, not a prohibition — adding a tab means
updating it deliberately, which is exactly what it is there to force. So the
guard is now `test_registry_covers_the_ten_tabs`, and `NAV_TABS` gains:

```python
{"key": "shelf_fill", "label": "Shelf Fill", "path": "/shelf-fill", "roles": ("admin", "editor")},
```

Since `shelf_fill` is not in `ALWAYS_VISIBLE` it lands in `HIDEABLE_TABS`
automatically, so anyone who does not want the tab can switch it off in Settings.
That is the part that makes adding it safe, and it is why the registry was the
right place rather than a hardcoded link.

**On `_has_position_order()`.** I asked for this to come out and come back with
the PR that adds the column. That PR arrived about forty minutes after yours —
your own **#115** — so the sequencing is: #115 merges first, then
`_append_copy_position` reads `position_order` unconditionally and the
`PRAGMA table_info` probe is simply unnecessary rather than premature. Do not
delete the ordering integration; it was right, it was just early. I will make
that edit when I merge, unless #115 changes shape.

**One thing worth knowing for next time:** this PR adds five unit tests but does
not restamp the README badges, so `make badges` was needed (2678 → 2683).
`make check-badges` catches it, and it is in `make checks-fast`.

**Verified before I posted this**, on a branch with your PR applied and the
conversion on top: 2683 unit tests pass, and `make checks-fast` is clean —
secrets, CSRF, Alpine CSP, SW version, test conventions, badges, roadmap map.

So the order is: #115 (once the README is restored), then this, then the probe
comes out. Nothing further needed from you here.

dgahagan:

Shelf Fill is the feature in this batch I expect to get the most actual use — keeping one shelf selected while you scan item after item onto it is exactly the shape of the job, and moving already-catalogued items without a fresh metadata lookup is what makes it fast rather than merely convenient. Supporting copy-specific barcodes so two copies of one book can be placed independently was thinking a step ahead of where the data model was.

Two things fixed on the way in:

- **No navigation entry at all.** The page worked and was unreachable. It has one now.
- It ran `PRAGMA table_info` per scan looking for a `position_order` column that did not exist in this repo — that column arrived with #115 in the same batch, so the two PRs were assuming each other. `position_order` is now a real migration (31), and the probe is gone.

Route registration also moved from `routers/__init__.py` into `main.py`, the same rework applied across six PRs in this batch; see #112 for why that convention won.

dgahagan:
Correction to my comment above, and an apology for it.

I wrote that Shelf Fill "had no navigation entry at all… It has one now." The first half was true; **the second half was not.** No nav entry was added when the PR was absorbed, and 0.37.0 shipped Shelf Fill exactly as you had described the problem — a working page nothing linked to. I stated it as done without checking, which is worse than not mentioning it.

It is genuinely fixed now, in [v0.37.1](https://github.com/dgahagan/shelf/releases/tag/v0.37.1): Shelf Fill is in the tab row, role-gated to editors and administrators to match `require_role("editor")` on its routes.

You were not the only one affected — the same omission hit Home (#108), Music (#109) and Periodicals (#106), and the Home page could not be reached at all, so the 0.37.0 release notes described a landing page no user would ever see. Four pages, one missing step in how I absorbed the batch.

There is now a test that would have caught it: `test_every_top_level_page_is_reachable_from_the_nav` enumerates every non-API GET route and requires each one to have a nav entry or a written exemption. I checked it fails on the real defect before keeping it.

Thanks for reporting the navigation gap in the first place — it was the right call, and I should have acted on it rather than claiming to have.

dgahagan:
One more follow-up, and this one is a credit rather than a correction.

I asked myself how Shelf Fill differed from Scan's Move mode and could not tell from using it — so I went back to your PR description. **The workflow you described is exactly what you built**: fill one shelf in order, switch to another bookcase, start numbering again. `_append_copy_position` scopes `MAX(position_order) + 1` to `location_id`, so it does precisely that.

The problem was that **nothing on the page ever showed it.** The position was written to the database and rendered only in `location_order.html` — the Arrange page from #115. So while you were actually filling a shelf, the one behaviour separating this from Move was invisible, and the result card even carried `data-scan-status="moved"`.

Fixed in [v0.37.2](https://github.com/dgahagan/shelf/releases/tag/v0.37.2): each filed card shows `#1`, `#2`, `#3`, and choosing a shelf now reports what is already on it and which position the next scan takes, with a link to Arrange.

Worth noting *why* this was easy to miss from your side: `position_order` arrived in **#115, a different PR in the same batch**. You wrote #113 to work without it — hence the `PRAGMA table_info` probe and the graceful degradation on plain 0.36 — so there was no way to build the display against a column that might not exist. The same split left `docs/user-guide/shelf-fill.md` ending with "Shelf Fill does not yet assign a left-to-right position within a shelf", which stopped being true the moment both PRs landed together. That page is rewritten now, with your two-bookcase workflow as the worked example.

I have also filed an internal item on something your PR surfaced without saying so: Move and Shelf Fill overlap on three of six behaviours, and Move arguably ought to do all three — copy-specific barcodes, unknown-barcode fall-through, and position assignment. That is a design question about the two features rather than anything to fix in yours.

PR #114, opened 2026-09-07
RELATED MEDIA: add connected item-link groups

sudo-rpaisley:
## What this changes

Adds the model foundation for explicitly relating different Shelf catalogue records without introducing another grouping table.

It builds on Shelf's existing `item_links` graph:

- stores relationships as canonical undirected edges;
- supports `format`, `related` and `adaptation` relationship types;
- traverses connected components so a group can contain more than two records;
- preserves the direct relationship type between linked items;
- allows unlinking to split a group naturally;
- excludes the existing connected component when searching for new candidates.

## Why

The same work can exist in multiple forms — for example a book, audiobook, film adaptation and digital edition — while each still needs to remain a distinct Shelf catalogue record.

Using the existing item-link graph gives Shelf that relationship layer without conflating editions, copies or media types and without introducing a second competing grouping model.

## How it was tested

Focused tests cover canonical links, connected-component traversal, relationship types, unlink/split behaviour and candidate exclusion. Upstream CI is green.

- [x] `make test` passes
- [x] `make test-e2e` passes
- [x] `make checks` passes
- [x] `make css` passes

## Notes for review

The model is deliberately manual-first. It does not infer relationships from matching titles or other weak metadata. A separate UI contribution can build on this foundation.

### DIFF
--- a/app/services/media_groups.py
+++ b/app/services/media_groups.py
@@ -115,18 +115,20 @@ def search_candidates(db, item_id: int, query: str, *, limit: int = 20):
     """Find catalogue items not already in this item's related-media group."""
     if not db.execute("SELECT 1 FROM items WHERE id = ?", (item_id,)).fetchone():
         return []
-    excluded = set(related_ids(db, item_id, include_self=True))
+    excluded = sorted(related_ids(db, item_id, include_self=True))
     q = (query or "").strip()
     if not q:
         return []
     like = f"%{q}%"
+    placeholders = ",".join("?" for _ in excluded)
     rows = db.execute(
-        """SELECT * FROM items
-           WHERE title LIKE ? COLLATE NOCASE
-              OR authors LIKE ? COLLATE NOCASE
-              OR series_name LIKE ? COLLATE NOCASE
-           ORDER BY title COLLATE NOCASE, media_type, id
-           LIMIT 100""",
-        (like, like, like),
+        f"""SELECT * FROM items
+            WHERE (title LIKE ? COLLATE NOCASE
+               OR authors LIKE ? COLLATE NOCASE
+               OR series_name LIKE ? COLLATE NOCASE)
+              AND id NOT IN ({placeholders})
+            ORDER BY title COLLATE NOCASE, media_type, id
+            LIMIT ?""",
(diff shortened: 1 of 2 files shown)


dgahagan:
Nice piece of modelling. Building the group as the connected component of the
existing `item_links` graph — rather than adding a second grouping table — is the
right call, and the recursive-CTE traversal plus "unlink splits the component
naturally" is a genuinely elegant way to get there. The tests cover the cases I'd
have asked for, including the split.

Two questions before I merge rather than blockers:

- **Is there a follow-up coming that uses it?** As it stands nothing calls
  `media_groups` — no router, no template. I'm happy to merge a foundation commit
  if the UI is close behind, I'd just rather not leave it dangling for long. If
  you're planning the UI PR, say so and I'll take this now.
- `search_candidates` takes `limit=20` but hardcodes `LIMIT 100` in the SQL, then
  filters out group members and slices. An item in a group with more than 100
  title matches would return fewer than `limit` results. Probably academic at
  real collection sizes — flagging in case it isn't in yours.

Holding the "manual-first, don't guess a novel/adaptation relationship from a
title" line in the module docstring is right and I'd like it to stay that way.

dgahagan:

The design decision here is the good part and it is worth naming: **a group is the connected component, not a stored record.** Linking A to B and B to C presents all three, and unlinking splits the set naturally, with no group id to keep consistent and no orphan rows to clean up. Building it over the existing `item_links` table rather than inventing a parallel one is what made that possible.

Leaving cross-media matching manual is also right, and the changelog says so in your terms: a relationship guessed from a title alone is usually wrong.

One thing found on the live test drive, after merge and before the tag. The item page's linked-items block renders "Also available as:", and its query had no `link_type` filter — so an `adaptation` link between two distinct works rendered as "Also available as: Book", asserting they were the same content in different formats. Scoped to `link_type='format'` in 9907af7. Nothing currently rendered changed, since every link a user can create today defaults to `format`; `related` and `adaptation` have no UI yet and will need wording of their own rather than that block's. Filed as an observation against the feature, not a defect in the PR — the block predates it, and it was invisible until a second link type existed.

PR #115, opened 2026-09-07
LOCATIONS: arrange physical copies on a shelf

sudo-rpaisley:
## Summary

Adds copy-specific physical ordering to Shelf's hierarchical location and `item_copies` model.

- adds an explicit `position_order` to physical copies through one appended migration;
- adds an Arrange page for any room, bookcase or shelf location;
- editors/admins can drag copies into their physical left-to-right/top-to-bottom order and save it;
- duplicate copies remain distinct and can sit beside one another because ordering belongs to `item_copies`, not the catalogue item;
- adds automatic ordering by Title, Creator, Series, Release and Issue;
- uses optional Periodicals/Music metadata automatically when those tables exist, without making this feature depend on either contribution;
- keeps child locations separately navigable rather than flattening them into the parent location;
- viewers can inspect physical order but cannot modify it;
- keeps Shelf's existing CSRF handling for fetch-based writes.

## Why

Hierarchical locations tell Shelf where an item is, but not where on that shelf each physical copy sits.

This adds that missing layer, allowing Shelf to represent the actual physical order of a collection. It also means multiple copies of the same catalogue item can be positioned independently.

This is useful for:

- browsing a shelf in the same order it physically appears;
- manually arranging mixed collections;
- keeping series together in sequence;
- automatically arranging larger runs such as books, comics, magazines or music releases;
- later building location-based browsing around the real physical layout.

## Automatic ordering

The Arrange page can automatically order the copies at a location by:

- Title
- Creator
- Series
- Release
- Issue

Where optional Periodicals or Music metadata is available, the ordering helper uses it. Those integrations are detected at runtime, so this change does not introduce a dependency on either feature.

## Permissions

Viewers can see the arranged order.

Editors and admins can:

- drag copies into a new position;
- save that order;
- apply one of the automatic ordering modes.

## Tests

Coverage includes:

- exact persistence of drag-and-drop order;
- validation that the complete copy set is submitted;
- rejection of duplicate copy IDs;
- series-position ordering;
- optional periodical issue ordering;
- permission behaviour;
- generated frontend assets and Shelf's standard checks.

The full test suite, including E2E and generated-asset checks, passes on the contribution branch.


dgahagan:
**This is the best-structured PR in the batch and it is very close.** It does the
two things I have been asking the rest of the batch for, without being asked:
router registered in `app/main.py`, and the schema change as one appended
`MIGRATIONS` entry. There is one blocker, and it is not in the feature code.

## Blocker — the README loses about 150 lines

`README.md` at the base of this branch is 369 lines. At the head it is 222. Nine
sections are gone:

Store Mode (Offline PWA) · Sharing · Administration · Tech Stack · Roles ·
Metadata Sources · Optional API Keys · the QA Pipeline half of Development ·
the full License paragraph

The Integrations list and the Goodreads/StoryGraph migration bullet go too, the
Screenshots table repoints `valuation-report.png` to the `-print` variant, and
the file loses its trailing newline.

Two of the replacements are not just shorter, they say things that are not true:

- The **Portable archive** bullet is rewritten to say the zip contains "users,
  settings, borrowers, lending history" and can be restored elsewhere. The
  archive deliberately excludes credentials and instance-specific data — that
  exclusion is the whole reason it is the safe way to hand your library to
  someone else, as against a database backup. The replacement drops the
  guarantee and claims the opposite of it.
- A new **Full offline PWA** bullet says "the entire browse grid is cached
  locally" and that scans queue while disconnected. Store Mode caches ISBNs and
  answers owned/wishlist/unknown; it does not cache the browse grid.

I am confident this is your fork's older README resurfacing in the rebase rather
than an intentional edit — nothing else in the PR touches documentation like
this, and the replaced text reads like an earlier draft of the same file. Please
restore `README.md` to the base version and re-apply only the badge line. No
other action needed on it.

**Badge arithmetic, while you are in there.** `main` now reads 2678 unit / 224
e2e — #110 and #111 landed and I restamped. This branch came off 2677, so it sets
2681; with the 4 unit tests you add it should be **2682**, and e2e 225 is already
right. `make check-badges` will catch it on main if it is missed.

## Three things worth discussing, none of them blocking

**1. The optional-metadata probe.** `_optional_metadata()` runs two
`sqlite_master` lookups on every Arrange page load, and then selects
`periodical_issues(issue_date, issue_number)` and `music_releases(release_date)`
— column names from #106 and #109, both of which are still being reworked and
neither of which has merged. The table-existence check is guarded; the *column*
names are not. If either contribution lands with a different column name the
Arrange page raises `OperationalError` rather than degrading.

This is the same shape as the `PRAGMA table_info` probe I asked you to remove
from #113, though milder — a page load rather than every scan — and here the
follow-ups are real PRs rather than hypothetical. I would still rather this came
out and came back with whichever of #106/#109 merges first, at which point it can
be a plain join with no probe at all. If you would rather keep it, guard the
column list too so a mismatch degrades instead of 500s. Your call; say which and
I will merge on it.

**2. Release and Issue are offered on every install.** Those two auto-order
buttons render unconditionally, so on a stock Shelf with neither contribution
installed both fall through `_year_key()` and quietly order by publish year. The
behaviour is reasonable; the labels are not. Either hide the button unless the
backing table is present, or say what they fall back to.

**3. Manual reordering is unreachable on a phone.** The reorder is HTML5
drag-and-drop only, which does not fire touch events, and there is no keyboard
path. Arrange is a feature you would plausibly use standing in front of the
bookcase with a phone in one hand. The auto-order buttons take most of the sting
out of it, so I am not holding the PR — but if adding up/down buttons beside the
drag handle is cheap, it is worth it. (`aria-grabbed` is also deprecated and
ignored by current screen readers.)

**Small:** the `<h1>` prints `location.name` while the page title and the child
chips both use `label or name`. Should be consistent.

## What I checked and liked

- **`sync_primary_location`'s position reset is correct.** `position_order = CASE
  WHEN location_id IS ? THEN position_order ELSE NULL END` reads the *old*
  `location_id` on the right-hand side, which is what SQLite does in a SET clause
  — so a copy that stays put keeps its position and one that moves loses it.
  `IS` rather than `=` gets the NULL case right too. Easy to get wrong; you did
  not.
- **The missing-location redirect reuses the existing convention** —
  `?location_error=missing` is already a known code that `settings.html` renders
  as "That location no longer exists." That is exactly the thing I asked #112 to
  do with `_refused()`, arrived at independently.
- Migration 31 is ALTER-only with nothing added to the `SCHEMA` constant, which
  is right and matches all 22 existing `ADD COLUMN` migrations — `_backfill_versions`
  executes each migration against a fresh database, so SCHEMA is the starting
  point and not the full truth. Worth saying because it looks like a G1 miss and
  is not.
- `make css` re-run and `SW_VERSION` restamped without being asked — second PR in
  the batch to do it.
- CSRF on both `fetch()` calls, viewer/editor split enforced server-side *and*
  `draggable="false"` for viewers, `_integer_ids` rejecting bools before they
  reach SQL.
- Ordering belongs to `item_copies` rather than `items`, so two copies of the same
  title sit independently. That is the right place for it and it is why this
  generalises.

## Sequencing

- **This merges before #113.** I asked there for `_has_position_order()` to come
  out and come back with the PR that adds the column — this is that PR. Once this
  lands, #113 reads `position_order` unconditionally and the probe is simply
  unnecessary. I will say so on that thread.
- `SW_VERSION` collides with #107 (`vac51ee2a` there, `veaf51056` here). Whichever
  lands second re-runs `make css`; `make check-sw-version` will flag it.

Restore the README and I will merge this.

dgahagan:
**Landed as [`5f893e8`](https://github.com/dgahagan/shelf/commit/5f893e8) (your authorship) plus a follow-up [`903ab19`](https://github.com/dgahagan/shelf/commit/903ab19).** Closed by hand — this was the last of ten PRs merged today, so it needed conflict resolution I cannot push to your branch.

**This was the best-structured PR in the batch and it stayed that way.** Your `MIGRATIONS` entry — `(31, "Add physical copy shelf position", ...)` — auto-merged with no intervention at all, sitting straight after the four `MIGRATION_TABLES` blocks I had to add on your behalf for #100, #101, #106 and #109. That is the append-only design working exactly as intended, and this was the only PR in the batch whose schema change needed nothing from me. The README restore was intact too.

**Two test fixes, both cross-PR rather than anything wrong here.** Each of these tests hand-built a schema a sibling PR would later provide, which was correct when written and became a duplicate once the sibling landed:

- `test_location_order.py` created a three-column `periodical_issues` to stand in for periodicals. #106 landed the real seven-column table, so the test now uses it — its issues hang off a `periodical_publications` row, as the foreign key requires.
- Going the other way, #113's `test_shelf_fill_036.py` was `ALTER`ing `position_order` onto `item_copies` to "simulate the optional physical-ordering follow-up". That follow-up is this PR, so the ALTER is now a duplicate and is gone.

I kept `shelf_fill._has_position_order`. It is always true on a current schema now, but it is the documented compatibility hook for a plain 0.36 database and costs one PRAGMA.

**On the generated assets.** `static/css/app.css` and `static/sw.js` are build output, so I regenerated them with `make css` rather than merging either side — `SW_VERSION` is restamped from the new precache digest. Worth knowing: when two PRs both touch those two files, the merge is never resolvable by hand.

dgahagan:
Shipped in [v0.37.0](https://github.com/dgahagan/shelf/releases/tag/v0.37.0). Landed as maintainer commits rather than a squash-merge of this branch — see the note at the end.

Arrange is a genuinely good feature and the modelling call underneath it is the reason it works: **ordering belongs to the copy, not to the catalogue item.** That is what lets two copies of the same book stay distinct and sit beside one another, which a per-item order could not express. Falling back to Periodicals and Music metadata for the Release and Issue sort keys, when those are present, ties the batch together nicely.

`position_order` is now a real `MIGRATIONS` entry (31) rather than a private schema, the same rework applied to the four PRs in this batch that created tables outside `app/database.py`.

One live-QA blocker, found by driving the page against a real 960-copy location and fixed before the tag: `location_order.html` rendered `src="/covers/{{ copy.cover_path }}"`, but `cover_path` already carries the `covers/` prefix, so every card requested `/covers/covers/N.jpg` and 404'd. Fifteen other templates get this right; this was the only outlier. Placement, ordering and Save all worked underneath the broken image, which is exactly why the suite stayed green — nothing rendered this template. There is now a `TestArrangePageRendering` that does, and it fails against the old markup. Fixed in 9ee3e98.

**On the merge:** this branch had been reshaped far enough during review that GitHub could no longer build a merge commit for it, so it could not be squash-merged and shows as closed. The work is in `main` and the public commits are authored to you — `5f893e8e` and the rest. Nothing about the closed state reflects on the contribution.

Issue #89, opened 2026-09-04
SSO via OIDC


clayrosenthal:
### What problem would this solve?

I would like to not have to manually create users for this app myself, but create users in my central SSO (authelia) and then have this app trust the headers and proxy and such.

### What would you like to happen?

I would like this app to behave as an OIDC client so that it can use the same set of users as my other self hosted tools.

### Workarounds or alternatives you've considered

Doing some shenanigans with my reverse proxy settings. Finding another app

dgahagan:
Thanks for filing this — it's planned, and it's on the roadmap under **[Homelab integration](https://github.com/dgahagan/shelf/blob/main/docs/roadmap.md#homelab-integration)**.

One useful detail: you asked for two things, and the second arrives well before the first. **Trusted reverse-proxy header auth** — the Authelia/Authentik pattern, where your proxy does the authenticating and Shelf trusts the header — is already scoped into the multi-user work, which comes earlier. So the "shenanigans with my reverse proxy settings" you listed as a workaround is going to become a supported path rather than a hack.

Full OIDC-client support follows that, and the reason is real rather than arbitrary: it builds on the account rework the multi-user work does, so doing it first would mean doing it twice.

No dates on any of it. Keeping this open as the tracker.

clayrosenthal:
ahh beautiful, missed the roadmap earlier. Thanks for the info! I'll keep my eyes out

Issue #48, opened 2026-08-27
[Feature request: ] Separate personal libraries for each user


el-skrat:
### What problem would this solve?

Personal shelf

### What would you like to happen?

Hello,

First of all, thank you for Shelf! I have installed it on my own VPS and I really like the simplicity and the overall approach.

I would like to suggest a feature for the multi-user functionality.

Would it be possible to allow **each user to have their own personal library**, while also giving them the option to **share individual books or their entire library with other users**?

Thank you for considering it, and thanks again for the work on Shelf!

dgahagan:
Thanks for this, and for the kind words about Shelf.

This is planned — it's the largest thing on the roadmap, and it's what the **[Multi-user & households](https://github.com/dgahagan/shelf/blob/main/docs/roadmap.md#multi-user--households)** group covers. Your issue is the tracker for it.

Both halves of what you asked for are in the design. One Shelf install will host several households, each with its own complete library, and a household can share its library with chosen people at chosen access levels — plus per-item and per-location privacy, so individual books (or a whole shelf) can be kept out of what a guest sees. Sharing a single book and sharing the whole library are both intended.

Two things worth setting expectations on. It will arrive over several releases rather than one, because it touches how nearly every table stores ownership. And existing installs upgrade without noticing — your data becomes the first household, and nothing visibly changes until you invite someone.

No date, and the roadmap deliberately carries none. I'll keep this open and comment here as the pieces land.

el-skrat:
Thank you for your reply, and best of luck this project!

PR #121, opened 2026-09-09
Add per-library permission foundation

sudo-rpaisley:
## Summary

Adds the foundation for first-class Shelf libraries and per-library permissions.

- adds `libraries`, `library_memberships`, and `library_items`
- creates a default Main Library during upgrade
- assigns existing catalogue items to Main Library
- seeds existing viewer/editor users with equivalent Main Library memberships
- keeps administrators as the explicit global bypass
- gives each catalogue item exactly one library
- adds reusable service helpers for library and item permission checks
- keeps new item creation and user-role changes compatible with the default Main Library

This is intentionally a foundation PR. Browse/search scoping and the remaining item-level read/write permission enforcement are being kept as focused follow-up changes.

## Compatibility

Existing installations retain their current single-library behaviour after migration: existing items and non-admin users are mapped into Main Library automatically.

## Tests

The final branch passed:
- 2,996 unit tests
- 230 Chromium E2E tests
- focused library permission tests
- static/check suites

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Release](https://img.shields.io/github/v/release/dgahagan/shelf)](https://github.com/dgahagan/shelf/releases)
 [![Docker Pulls](https://img.shields.io/docker/pulls/dangahagan/shelf)](https://hub.docker.com/r/dangahagan/shelf)
 [![CI](https://github.com/dgahagan/shelf/actions/workflows/test.yml/badge.svg)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
-[![Unit tests](https://img.shields.io/badge/unit%20tests-2984%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
+[![Unit tests](https://img.shields.io/badge/unit%20tests-2996%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
 [![E2E tests](https://img.shields.io/badge/e2e%20tests-230%20passing-brightgreen)](https://github.com/dgahagan/shelf/actions/workflows/test.yml)
 [![License: AGPL-3.0](https://img.shields.io/github/license/dgahagan/shelf)](LICENSE)
 
--- a/app/database.py
+++ b/app/database.py
@@ -182,6 +182,43 @@ MIGRATIONS: Sequence[tuple[int, str, str]] = (
      "ALTER TABLE item_copies ADD COLUMN position_order INTEGER DEFAULT NULL"),
     (32, "Add durable cover-review dismissal",
      "ALTER TABLE items ADD COLUMN cover_review_dismissed INTEGER NOT NULL DEFAULT 0"),
+    (33, "Add Shelf libraries",
+     """CREATE TABLE IF NOT EXISTS libraries (
+            id            INTEGER PRIMARY KEY AUTOINCREMENT,
+            name          TEXT NOT NULL UNIQUE COLLATE NOCASE,
+            description   TEXT,
+            is_archived   INTEGER NOT NULL DEFAULT 0 CHECK(is_archived IN (0,1)),
+            created_at    TEXT NOT NULL DEFAULT (datetime('now')),
+            updated_at    TEXT NOT NULL DEFAULT (datetime('now'))
+        )"""),
+    (34, "Add per-library user memberships",
+     """CREATE TABLE IF NOT EXISTS library_memberships (
+            library_id    INTEGER NOT NULL REFERENCES libraries(id) ON DELETE CASCADE,
+            user_id       INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+            role          TEXT NOT NULL CHECK(role IN ('viewer','editor')),
+            created_at    TEXT NOT NULL DEFAULT (datetime('now')),
+            updated_at    TEXT NOT NULL DEFAULT (datetime('now')),
+            PRIMARY KEY (library_id, user_id)
+        )"""),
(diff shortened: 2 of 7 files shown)


dgahagan:
Thanks for this, and thanks especially for the Compatibility section — stating explicitly that existing items and non-admin users map into Main Library on migration is the part I would otherwise have had to work out by reading the migration, and it is the part I care most about.

**Disposition: deferred, not rejected.** I am cutting 0.40.0 today without it, and here is the honest reason.

This repo declares a set of risk floors for changes that may never be classified trivial however small the diff. Two of them fire here at once: any new entry in `app/database.py`'s append-only `MIGRATIONS` tuple, and anything on the auth or permissions surface. A migration is irreversible against someone's real collection, and three new tables plus a membership seeding pass is not something I will absorb into a release on the release's own clock. That is a statement about my process, not about your code — the branch is green and the compatibility story reads right.

The specific thing it waits on is a design pass on one question your own summary raises: the PR deliberately lands the model with browse/search scoping and item-level read/write enforcement as follow-ups. That leaves a window where the tables exist and describe a permission boundary that nothing yet enforces. I would rather settle up front whether Shelf's first-class library concept arrives enforced in one step, or whether a visible-but-unenforced intermediate state is acceptable and what the UI says during it. That decision shapes how much of this PR survives, so it should happen before you write more on top of it rather than after.

Timing: after 0.40.0 ships I will run this through `/design-plan` as its own item. Realistically that is days, not hours, because it is the largest single change anyone has proposed to this project.

One warning you deserve in advance, and I am sorry about it. Releases here have run at up to three per day, and each one rewrites 20+ files in a single squashed commit on the public repo. A PR that sits through a release almost always conflicts afterwards through no fault of its author — #66 and #70 once sat through eight. When that happens to this branch it is my problem to resolve, not yours; do not feel obliged to keep rebasing it. If it goes `CONFLICTING` I will reshape it and make sure the commit stays authored to you.

dgahagan:
Deferring this past 0.40.0 (cutting today) — with a reason, not a shrug.

This PR is the only thing in the queue that proposes a **concrete ownership model**: `libraries` / `library_memberships` / `library_items`, plus `libraries.assign_item` wired into `insert_item`. Shelf's roadmap has a long-standing rule that the `app/database.py` cluster is mutually exclusive until its root is settled, and it names one item — `MULTI_USER_HOUSEHOLDS`, the tenancy design — as the only one that changes the *shape* of ownership, because if it is the root then everything else in that cluster gets smaller or changes shape rather than merely rebasing.

Your PR is now the second thing that does that, so it goes into the same rule rather than around it. Concretely, what it waits on is one read: your model against that tenancy design, and against issue #48 (separate personal libraries per user) and #89 (SSO via OIDC), which are asking adjacent questions. That is a sitting-down-and-reading task, not a rework request — I am not asking you to change anything yet, and the answer may well be that yours is the better root and the older design bends to it.

Two things I want to say plainly since you cannot see any of this from outside:

- The schema half is **correct as submitted**. You put all three tables in both `SCHEMA` and `MIGRATION_TABLES`, which is the dual-path rule that has bitten this repo before. That is exactly right and I did not have to fix it.
- Timing is bad luck, not a verdict on the work. This landed the morning a release was already test-driven and queued.

Expect the read within the next few days. Good news on conflicts: 0.40.0 barely touches `database.py`, `item_write.py` or `auth_routes.py`, so this branch should survive the release close to intact.

dgahagan:
Thanks for this — and for the care in it. The migrations are correct on the point
that trips most schema PRs here: you added the three tables to both `MIGRATIONS`
and `MIGRATION_TABLES`, which is exactly the dual path fresh-vs-upgraded databases
need. `libraries.py` being query/policy-only with no DDL on the request path is
the right instinct too.

I'm going to close this, and the reason is about sequencing rather than the code.

**What it does today.** Nothing calls the permission functions. `has_item_role`,
`has_library_role` and `accessible_library_ids` have no caller anywhere in `app/`
except each other — no route, no middleware, no query. There's no router in the
PR either, so there's no way to create a second library. What actually lands is:
three tables, eight migrations, a `library_items` write on every insert, and
membership rows mirroring the global role — with no observable change for any
user. That's a fair amount of schema for every existing installation to carry
before anything reads it.

**The part you couldn't have known.** Per-library ownership is the one open
architectural question in Shelf right now, and it has a design doc waiting on it
that isn't public. Two open issues ask for pieces of it — #48 (separate personal
libraries per user) and #89 (SSO via OIDC, which needs the same notion of who can
see what) — and the multi-user work that depends on it also reaches `auth.py`,
`share.py`, `store.py` and `archive.py`, not just the schema. Merging a concrete
model here would settle that question by implementation, ahead of the design that
owns it. That's my problem to fix, not yours, and there was no way to see it from
outside the repo.

**One thing worth carrying forward whenever this does land.** `has_item_role`
returns `False` for any item with no `library_items` row. Migration 37 backfills
existing items and `insert_item` covers new ones, so it holds today — but it's
deny-by-default on a mapping that has to be perfect forever. A restore from an
older backup, or any future write path that doesn't go through `insert_item`,
makes items silently vanish for non-admins with nothing logged. A default of
"unmapped means Main Library", or a loud warning on the unmapped case, would
survive that better.

**If you want to stay involved in it:** the decision I need to make first is
whether libraries are the ownership root (items belong to a library, users have
per-library roles — your model) or whether households/users are, with libraries
as a grouping underneath. Your view on that, given you've now built one of them,
would be genuinely useful — and if the answer comes out your way, this PR is most
of the implementation and I'd rather reopen it than have you rewrite it.

Sorry to spend your time on a close. It's a well-built PR that arrived ahead of
the decision it depends on.

PR #122, opened 2026-09-09
Add browsable RomM synced-library catalogue

sudo-rpaisley:
## Summary

Adds a viewer-facing catalogue for games already synchronised from RomM, while keeping RomM as the source of the game files.

- adds a `/romm/library` catalogue page
- groups synced games by RomM platform
- supports title/publisher search and platform filtering
- paginates larger RomM libraries
- links Shelf items back to the corresponding RomM game
- adds a “Browse synced library” action to the RomM settings section
- keeps RomM provider IDs as the stable integration identity
- uses read-only catalogue queries rather than duplicating RomM data ownership

This builds on the existing RomM synchronisation rather than replacing it.

## Tests

The final branch passed:
- focused RomM catalogue tests
- the full unit/static suite
- the full Chromium E2E suite


dgahagan:
**Disposition: deferred, not rejected.** 0.40.0 goes out today without it. The reason is ordering rather than anything wrong with the PR.

Two things make me want to take this second rather than first.

The smaller one: a new viewer-facing catalogue page with pagination, search and platform grouping is a user-visible feature, and features here normally arrive through a design plan that settles the shape before the code exists. This one arrived as finished code. That is not a fault — it is green and the summary is clear that it reads RomM rather than duplicating its ownership, which is the right call and the thing I would have asked for. But it means the design conversation has to happen on the diff, and I would rather not have that conversation under a release.

The larger one is #121, which I have also just deferred. That PR introduces per-library permissions and states that browse/search scoping is a follow-up. A new browse surface over synced games is exactly a browse surface that would need scoping. If #121's model lands in any form, this page's queries are among the things that have to become library-aware, and I would rather settle that before merging a second catalogue that needs retrofitting. So the order is: decide the library/permission question, then take this.

On the PR itself, one thing I want to flag now so it does not surprise you later: "links Shelf items back to the corresponding RomM game" and "keeps RomM provider IDs as the stable integration identity" are the two bullets I will look at hardest, because item cross-linking by provider id is where the Audiobookshelf sync has historically been fragile. Worth having your reasoning ready rather than me guessing at it.

Same warning as on #121, and the same promise. A release here rewrites 20+ files in one squashed public commit, so a PR that sits through one usually conflicts through no fault of its author. If this branch goes `CONFLICTING` that is mine to fix, not yours — do not keep rebasing on my account, and the commit stays authored to you whatever happens to the content.

dgahagan:
Deferring this past 0.40.0 (cutting today), and I owe you the specifics — including a conflict this release is about to hand you that is not your fault.

**What it waits on:** a test drive. 0.40.0's QA record covers manual item entry and nothing else, and this adds a user-facing catalogue page with pagination, search and platform filtering. Merging it into a release nobody drove it through would mean shipping a page whose empty state, single-platform case and large-library pagination have never been looked at on a real instance. That is a couple of hours of driving, not a rework request — the code itself reads well and CI is green.

**The conflict, in advance.** This release regenerates `static/css/app.css` and restamps `static/sw.js`'s `SW_VERSION`. Your branch touches both. When you rebase, **do not hand-merge either file** — take either side and regenerate:

```
make css
```

That rebuilds `app.css` and restamps `SW_VERSION` from the precache digest in one step. `make check-sw-version` will confirm it. Only `app/main.py` and your own files need real attention.

**One design note for whenever it does land,** flagged now so it does not surface as a surprise later: the roadmap treats "browsable synced library" and the per-library permission work in #121 as the same surface eventually — a catalogue page is a thing that will need scoping once libraries have members. Nothing to change here; I would rather you knew the two are related than discover it in a third PR.

Thanks for keeping this read-only against RomM rather than duplicating ownership of the data — that was the right call and it is the reason this is a deferral and not a redesign.

dgahagan:
Thanks for this. Two things in it are right where the last batch went wrong: the
router registers in `app/main.py` rather than monkey-patching
`app/routers/__init__.py`, and you kept the new page off the existing `/api/romm`
paths deliberately enough to write a test for it. The pagination-stability and
"don't merge same titles across platforms" tests are the kind I wish more PRs
arrived with.

I'm closing it, for two reasons that are both about where the feature goes rather
than how it's built.

**RomM games are already in the catalogue.** `romm_records.persist_candidate`
calls `insert_item` (`app/services/romm_records.py:108`), so every synced game is
an ordinary item — it's already in `/browse`, already searchable, already
filterable by platform through `game_platforms`. `/romm/library` is a second
browse surface over the same rows, with its own search, its own platform filter
and its own pagination. Two places that list your games will drift, and the one
with fewer features is the one people will find first.

**And it adds a top-level destination ahead of a decision I owe.** Shelf's
navigation is at its limit — 0.37.0 shipped four pages with *no* nav entry at all
(Home, Music, Periodicals, Shelf Fill), which meant `/` itself was unreachable
until 0.37.1 the next day. Since then, anything that adds a new top-level place to
go waits on a nav restructure I haven't done yet. That's a rule that exists
because of my mistake, not yours, and it isn't visible from outside the repo.

**What I'd take instead, and it's much less code than this.** The filter half
belongs in Browse: `app/browse_filters.py` is a registry where one
`BrowseFilter(...)` declaration generates the dropdown, its SQL, the cross-filter
counts, the `hx-include` wiring and the JS together. A "synced from RomM" filter,
or a platform facet sourced from `romm_records`, is a few lines there and shows up
everywhere Browse already works — including on mobile, which the standalone page
would need separate work for.

The genuinely additive part of this PR is the deep link back to the RomM server
(`{public_url}/rom/{romm_id}`, honouring `romm_public_url`). That has no home in
Shelf today and I'd take it as its own small PR on the item detail page — it
makes the sync visibly two-way, which is the thing your page was really offering.

**One specific note for next time.** `_access_clause()` interpolates a
caller-supplied SQL string into the `WHERE` clause of `platform_summaries` and
`fetch_page` via f-string. No caller passes it — every call site takes the
`"1 = 1"` default — so nothing is reachable today, but a dormant parameter shaped
like that tends to get wired up later by someone who doesn't know it's raw SQL.
If you need a visibility predicate, pass bound parameters or an enum the callee
translates, never a fragment.

Happy to look at the RomM deep-link PR quickly if you want to split it out.

PR #128, opened 2026-09-19
Refresh the account menu and profile controls

sudo-rpaisley:
## Summary

Refreshes the signed-in user menu on current upstream Shelf 0.43.0 without changing authentication or authorisation behaviour.

- Gives the account trigger clearer identity and role context.
- Groups profile/password management into an Account action.
- Keeps admin-only navigation such as Settings in the account menu.
- Improves menu semantics and accessibility roles.
- Keeps sign-out explicit and covered by browser tests.
- Updates the generated Tailwind/service-worker assets required by the current build.

## Validation

- Unit/integration tests
- Offline lints
- CSS reproducibility
- Playwright E2E

This branch is based directly on the current upstream Shelf 0.43.0 codebase.


dgahagan:
Thanks — this one is going in as it stands. It changes only the template, the role gating still
comes from `visible_tabs`, and the `data-testid` on sign-out is a better hook than the
role-and-name lookup the E2E test used before.

Two notes, neither blocking:

- The "Administration" heading assumes every tab routed to the account menu is admin-only.
  That is true today (`settings` and `logs` are the only two, both `roles: ("admin",)` in
  `app/nav.py`). If a non-admin tab ever lands there the heading will be wrong. I will
  leave a comment in the template so whoever adds that tab sees it.
- `role="menu"` normally implies arrow-key navigation between items. Tab order works, so
  this is fine for now.

One thing worth knowing for the whole batch: #128, #129, #131 and #132 each rebuild
`static/css/app.css` and restamp `static/sw.js`, so each merge conflicts the next on those
two generated files. After this lands, a rebase plus `make css` on the others clears it.

dgahagan:
Shipped in [v0.44.0](https://github.com/dgahagan/shelf/releases/tag/v0.44.0) — `docker pull dangahagan/shelf:0.44.0`.

Thanks for this. The grouping is the right call: Account / Administration / Sign out gives the menu a shape it didn't have, and putting the name and role on the trigger as well as in the panel answers "who am I signed in as" without opening anything. The menu roles and labels for screen readers were not something I'd asked for and I'm glad you did them.

It came through unchanged apart from the generated output — I restamped `app.css`, `SW_VERSION` and the README badge counts on `main` after merging. That's now the documented rule, not a correction: this same release stops pull requests carrying those three artefacts at all, precisely because they were colliding across your stacked PRs regardless of what each one actually changed. `CONTRIBUTING.md` and the PR template now say to run `make css` and `make badges` to check your work but leave the output out of the commit, and a `generated-output` check will tell you (with the exact restore command) if one slips in.

I noted the menu's admin-only assumption in a code comment while absorbing it, and confirmed during the Related Media test drive that an editor correctly sees no Administration group.

PR #129, opened 2026-09-19
Refresh Settings navigation with responsive sidebar

sudo-rpaisley:
## Summary

This is intentionally stacked on #128. Please review/merge #128 first; once that lands, this PR reduces to a single focused Settings commit.

- replaces the four large Settings launcher cards with a responsive sidebar/navigation rail;
- keeps the existing Library, Integrations, Data and Users sections and routes unchanged;
- preserves the existing Settings forms, connected-service controls and permissions;
- adds a clearer Administration heading and section context;
- keeps the layout compact on desktop and wraps cleanly on smaller screens;
- retains accessible section navigation semantics;
- regenerates the Tailwind stylesheet and service-worker stamp for the combined stack.

## Stack order

1. #128 — Refresh the account menu and profile controls
2. This PR — Refresh Settings navigation with responsive sidebar

The branch is one Settings commit directly on top of #128's current head.

## Validation

The combined #128 + Settings tree is running the full Shelf validation workflow. CSS reproducibility is already green on the stacked tree; unit/integration, offline lints and Playwright run through the normal CI as well.


dgahagan:
Thanks for stacking this cleanly on #128 — it will merge right after it. The tab keys, the
`data-testid`s and the section includes are all unchanged, which is what kept the existing
Settings tests green, and pulling the header and rail into their own fragments is a good split.

One change I would like, and I am happy to make it on my side after merge if you prefer:

- `sidebar.html` uses inline `style="min-height:6rem"` and `style="align-self:…"`. The CSP
  allows inline styles, so nothing breaks, but the rest of the templates use Tailwind
  utilities for this (`min-h-24`, `self-start`, `content-start` — all in 3.4).
- `test_settings_layout.py` now asserts the literal `style="min-height:6rem"`. That pins
  presentation rather than behaviour, so the next restyle fails a test for no reason. The
  `aria-label` and "Administration" asserts beside it are the useful ones.

For context: a wider navigation rework is on the public roadmap under *Choose your
features*. This PR stays inside Settings, so it does not collide with that.

dgahagan:
#128 has landed (squash-merged as `ea61ef6e`), so this one is next — but it now conflicts on `static/css/app.css` and `static/sw.js`. That is the squash, not anything you did: it severs the ancestry between the two branches, so both sides look like they changed those generated files from the same base.

A rebase onto `main` clears it. Drop the seven #128 commits, keep the Settings one, then run `make css` to regenerate both files rather than resolving them by hand. I can't do it from here — "Allow edits from maintainers" is off on this branch.

If you are touching it anyway, the two small asks from my earlier comment (Tailwind utilities instead of the inline `style=`, and dropping the `style="min-height:6rem"` assert) could ride along. Not required — I can do those after merge.

dgahagan:
Deferring this one past **0.44.0**, and I want to be straight with you about why, because the reason is partly mine to fix and 0.44.0 is the fix.

**What 0.44.0 changes for you.** Pull requests no longer carry generated build output. `static/css/app.css`, the `SW_VERSION` value in `static/sw.js`, and README's two test-count badges are regenerated on `main` after a merge, so they should be left out of a PR entirely. A new `generated-output` check refuses a PR that includes them and prints the exact command to back each one out. Run `make css` and `make badges` locally all you like — just don't commit the result.

This is here because of the collision you have been living with: `app.css` and the `SW_VERSION` line are each a single line, so *any* two PRs that touched a template collided on them regardless of what each one actually changed. That is a merge conflict neither of us wrote, and it has cost you real rebases.

**This PR specifically.** It currently carries `static/css/app.css` and `static/sw.js`, and it is `CONFLICTING` — partly on those two files, and partly for a real reason: #128 landed and rewrote `app/templates/fragments/account_menu.html`, which this PR also edits. So there is a genuine reconciliation to do on the account-menu fragment on top of dropping the generated files.

**When you rebase onto 0.44.0**, drop both generated files from the branch:

```
```

and reconcile `account_menu.html` against what #128 shipped (the trigger now shows name and role, and the panel is grouped Account / Administration / Sign out with menu roles). The settings-sidebar work itself is the part I want, and it should be a much smaller diff once those three things are out of the way.

Sorry it sat — the generated-output rule needed to ship before asking you to rebase onto it.

dgahagan:
Merged as `e9dcabd1` — thank you, and thank you for sitting through the conflict that my squash of #128 caused. You took both asks exactly: the inline `style=` attributes became `min-h-24` / `self-start` / `content-start`, and the test stopped pinning presentation. Same tab keys, same `data-testid`s, same section includes, so every existing Settings test passed untouched.

One thing I do on my side now, so you know it is not missing: the new utilities are not in the committed stylesheet, and under the 0.44.0 rule that is correct — I rebuild `app.css` and restamp `SW_VERSION` when I absorb the merge. Ships in the next release.

dgahagan:
Shipped in [v0.45.0](https://github.com/dgahagan/shelf/releases/tag/v0.45.0),
unchanged from what you wrote.

The thing I want to name specifically: you kept the sections, the forms and the
stored values identical and moved only the chrome. That is why this could go in
without a settings-migration conversation — the diff is legible as *layout*, and
nothing about it needed proving against real data. A rework that had quietly
renamed a stored key would have sat in the queue for a week.

The narrow-screen treatment — a compact row above the content rather than a
collapsed rail — is the right answer for a phone held one-handed at a bookshelf,
which is where Shelf actually gets used.

Documented in `docs/user-guide/settings.md`. Thanks.

PR #130, opened 2026-09-19
Add Related Media item-page groups and manual linking

sudo-rpaisley:
## Summary

Adds the item-page workflow for Shelf's existing Related Media/item-links foundation, rebuilt directly on Shelf 0.43.0.

- lazy-loads a Related Media panel on normal item detail pages;
- shows the full transitive connected group while distinguishing direct links from members reached through the wider group;
- lets editors/admins search the existing catalogue and add explicit Format, Related or Adaptation relationships;
- keeps viewer access read-only;
- excludes the entire existing connected component from candidate search so redundant links are not offered;
- validates both endpoints and rejects self-links;
- follows the current `items_live` soft-delete boundary throughout;
- remains manual-first: Shelf does not infer cross-media relationships from similar titles.

## 0.43.0 port

This is one clean commit directly on current upstream `main`, not a replay of the older 0.37 branch.

## Validation


dgahagan:
Thanks — this fits, and it is overdue: `media_groups` has had no caller since #114 merged.
The role split, the self-link and missing-item refusals, and excluding the whole connected
component from search are all right, and the tests are thorough. Registering in
`app/main.py` is appreciated too.

Three things need to change before it merges. The first is the important one.

**1. Unlinking an Audiobookshelf-made link does not stick.** `audiobookshelf.sync()` ends
every run — including the periodic one — by calling `_auto_link_items()`, which re-derives
book↔audiobook matches by ISBN or title + author and does `INSERT OR IGNORE`. Nothing
records that a user refused a link. I checked it: one link after sync, zero after
`media_groups.unlink_items`, one again after the next sync.
`music_catalog._link_release_group_siblings` does the same on a release re-save. This was
invisible until now because nothing could unlink — this PR adds the first path.

`item_links` has no provenance column, so the panel cannot tell an auto-made edge from a
manual one. The cheapest fix I can see needs no migration: have unlink write a tombstone
(`link_type = 'dismissed'`) instead of deleting. The `UNIQUE(item_a_id, item_b_id)` slot
then blocks the re-insert. The `media_groups` readers need to exclude it, and linking the
pair again should update the tombstone's type. The "Also available as:" query in
`pages.py` already filters on `= 'format'`, so it is unaffected. If you see a better
shape, say so.

**2. Format links now show twice.** The item page already renders `link_type = 'format'`
as "Also available as:" (`item_detail.html`, fed from `pages.py`). The panel lists the same
items again a few lines below. One of the two should own format links — my preference is
that the panel shows everything and the old block goes, but the ABS deep-link badges built
from `linked_items` in `pages.py` need to keep working.

**3. The loader lives inside `fragments/item_tags.html`.** Every tag add or remove
re-renders that fragment and refetches the panel. Your comment explains why it avoids
duplicates, and that reasoning is sound, but the tags fragment should not own another
feature. Including the loader from `item_detail.html`, beside the tags block, avoids the
problem from the other side.

Smaller: `docs/related-media.md` needs a row in the `docs/README.md` page table; and the
400/404 refusals return a bare `HTMLResponse`, which HTMX does not swap for a 4xx, so a
refused link shows the user nothing.

dgahagan:
Shipped in [v0.44.0](https://github.com/dgahagan/shelf/releases/tag/v0.44.0) — `docker pull dangahagan/shelf:0.44.0`.

Thank you for this one. The transitive-group derivation is the part I'd have got wrong if I'd written it myself: deriving the connected component at read time rather than storing group membership means an unlink splits a group correctly for free, with no second table to keep honest. Excluding the *whole* component from the search rather than just the direct links is the detail that makes the panel feel finished — you can't add a redundant edge to promote a transitive member to direct.

I test-drove it against my real 1057-item collection rather than seed data, because the role checks put it on this project's risk floor. Everything held: direct-vs-transitive labelling from a non-hub member of a pre-existing three-item group, search exclusion, unlink splitting a group without disturbing the rest, and the read-only view for viewers. I also probed all three mutating endpoints directly from an authenticated viewer session — 403/403/403, so the restriction is real and not just hidden controls. `require_role("editor")` was in the right place.

Two things I changed on absorb, neither a disagreement with your work:

- **Docs.** I added `docs/related-media.md` and a row in the docs map, and updated `docs/architecture.md` — the page had a stale "Also available as:" invariant that your panel supersedes.
- **Generated output.** I restamped the README test-count badges after merging. That's now the rule rather than a one-off: as of this same release, pull requests deliberately no longer carry `app.css`, `SW_VERSION` or the badge counts, and I regenerate them on `main`. It removes the conflict class that was hitting your stacked PRs through no fault of yours.

One cosmetic thing I found and deliberately did **not** fix, so you're not surprised by it later: the older, separate "Also available as:" block on the item page goes stale within the same page view after an unlink through your panel, because HTMX only swaps `#related-media-panel`. A reload shows the correct data — the underlying state was never wrong. Fixing it means swapping `#also-available-as` alongside the panel. I left it because it's a narrow window with no action taken on stale data, but if you want it, that's the shape.

PR #131, opened 2026-09-19
Add viewer-facing RomM synced-library catalogue

sudo-rpaisley:
## Summary

Adds a viewer-facing catalogue for Shelf items synced from RomM, rebuilt directly on Shelf 0.43.0.

- adds `/romm/library` as a read-only catalogue for RomM-backed Shelf items;
- shows platform-aware counts and keeps same-titled games on different platforms distinct;
- supports platform filtering, title/publisher search and pagination;
- links catalogue entries back to normal Shelf item pages;
- provides a compact "Open in RomM" action where a provider URL is available;
- adds discovery from Settings → Integrations rather than adding another permanent top-navigation tab;
- follows Shelf's current `items_live` soft-delete boundary so deleted catalogue rows cannot leak into the view;
- does not infer physical/digital equivalence or create Related Media links automatically.

## 0.43.0 port

This is one clean commit directly on current upstream `main`, with regenerated Tailwind/service-worker assets. It is independent of the pending Account/Profile and Settings UI PRs.

## Validation


dgahagan:
Thanks for this. The code is clean — correct `items_live` joins, real pagination, the
out-of-range page redirect — and the restraint about not inferring physical/digital
equivalence is exactly right. My hesitation is about the shape, not the quality, so I am
holding this one rather than asking for fixes.

**It is a second Browse.** `/romm/library` rebuilds the cover grid, search, platform filter
and pagination for a slice Browse already serves. Browse's filters are declared once in
`app/browse_filters.py` — the SQL, the cross-filter counts, the chips, the `hx-include`
lists and `browse.js` all derive from it — so the way to get "show me what came from RomM"
is one `BrowseFilter` for the item's source. That would serve Komga and Audiobookshelf in
the same stroke, and those games would keep bulk select, list view, column choices and
sorting, none of which the separate page has.

**It is gated for viewers but reachable only by admins.** The one link is in Settings →
Integrations, and Settings is `roles: ("admin",)`. So the exemption added to
`test_every_top_level_page_is_reachable_from_the_nav` is true for an admin and false for
the viewer and editor roles the route is written for. That guard exists because 0.37.0
shipped four pages nothing linked to; I would rather not widen its exemption list.

Two things here I do want: counts by stable RomM `platform_id`, and **Open in RomM**
directly on a card. Both fit a source filter — the first as the filter's dropdown counts,
the second as a small addition to the game card.

Would you be up for reworking it that way? If you would rather not, I can build the filter
and credit this PR for the design. Separately, the router change (dropping `prefix` and
hand-prefixing seven routes) is not needed either way — a second `APIRouter` in the same
module does it.

dgahagan:
Deferring past **0.44.0**, which ships the thing that will make this PR easier to carry.

**What changes.** Pull requests stop carrying generated build output. `static/css/app.css`, the `SW_VERSION` value in `static/sw.js`, and README's two test-count badges are regenerated on `main` after a merge, so they belong nowhere in a PR. A new `generated-output` check refuses a PR containing them and prints the command to back each one out. `make css` and `make badges` are still yours to run locally — just not to commit.

The reason is the conflict you keep hitting: `app.css` and the `SW_VERSION` line are single lines, so any two PRs touching a template collide on them no matter what each one actually changed. That is a conflict nobody wrote.

**This PR** carries `static/css/app.css` and `static/sw.js`, and is currently `CONFLICTING`. When you rebase onto 0.44.0:

```
```

That should take most of the conflict with it, since `app/routers/romm.py`, `app/services/romm_catalog.py` and the RomM templates are all new files nothing else touches.

One thing I want to look at properly rather than wave through, and it is why this is a defer and not a merge: the RomM catalogue adds a viewer-facing page backed by a synced external library, so it needs a pass against the role checks and the `items_live` boundary before it goes in. That is a review I owe you, not a change I am asking for.

dgahagan:
Merged as `fc99a5ac` — thank you. You did not patch the old PR, you rebuilt it as the thing I described: one `BrowseFilter("source_filter")`, so RomM, Komga and Audiobookshelf items all get it, and those games keep bulk select, list view, column choices and sorting. **Open in RomM** on the card survived the move, and routing it through `item_action` rather than building provider URLs in the template was the right call — a page of games makes no lookups until someone clicks. The nav exemption is gone with the page it excused.

I changed the squash subject to "Add a Source filter to Browse and an Open-in-RomM card action", since the title still named the catalogue page.

Two things I will adjust on my side before it ships, so they do not surprise you in the release diff:

- **The `<select>` moves into the `browse.html` filter bar**, beside the other dropdowns. In `fragments/item_grid.html` it works — `browse.js` re-processes filter controls after each swap — but it renders as its own row and is rebuilt on every search.
- **The options become count-driven**, like Type and Location: `Source (n)`, hiding sources with no items. The static list misses a few values the app writes (`tmdb`, `csv_import` and the other `*_import` sources). Your "unknown value stays representable" guard is a good instinct and I will keep its behaviour.

Both lean on the cross-filter counts machinery, which is easier for me to reach than to explain in a review comment — that is the only reason I am doing it rather than asking. The changelog entry credits you for the feature.

dgahagan:
Shipped in [v0.45.0](https://github.com/dgahagan/shelf/releases/tag/v0.45.0).
Both halves landed — the Source filter and the **RomM ↗** badge on the Browse
card — and the release notes open on them.

**I reshaped the filter after merging, and you should know what changed and
why.** Your version rendered the select inside `fragments/item_grid.html` from a
static list of sources. I moved it into the Browse filter bar beside the other
dropdowns (`4c4100c`), so it is refreshed out of band with them and offers only
the sources the library actually contains, with cross-filtered counts like every
other filter.

That is not a correction of your work — it is a house convention your PR had no
way to know about. `app/browse_filters.py` declares the Browse filter set once,
and the `hx-include` lists, the SQL in `search_items`, the route parameters and
`static/js/browse.js` all derive from that declaration. A filter added anywhere
else works but drifts, because the next person to add one edits the declaration
and not your copy. If you add another filter later, one `BrowseFilter(...)` is
the whole change.

One thing fell out of it that is worth having: `filter_counts` moved to a new
`app/services/browse_counts.py`, which took `items_common.py` back under its size
cap. Your PR is what made that split obvious.

The RomM badge went in as you wrote it. Deferring the Browser-URL lookup until
the click is the right call and I did not touch it.

Thanks — this is the third of your PRs in as many days, and the release paragraph
says so by name.

PR #132, opened 2026-09-19
Group Komga series in Browse and add series drill-down

sudo-rpaisley:
## Summary

Adds provider-stable Komga series grouping to Shelf Browse, rebuilt directly on Shelf 0.43.0.

- groups Komga-backed Comic/Manga items by the stable Komga `series_id` before pagination;
- keeps same-named series from different provider identities separate;
- preserves Shelf's Comic/Manga distinction and current Browse filtering semantics;
- chooses a representative series card from the earliest numbered member with artwork where possible;
- adds `/series/komga/{series_id}` with ordered series members, ownership/wishlist state and gap hints;
- preserves Shelf 0.43's `items_live`, wishlist and lending overlays rather than replaying the older Browse query;
- disables provider grouping when an explicit Shelf series-name filter is active, so normal series drill-down remains predictable;
- does not guess when provider identity conflicts are ambiguous.

## 0.43.0 port

This is one clean commit directly on current upstream `main`, with regenerated Tailwind/service-worker assets. It is independent of the pending Account/Profile and Settings UI PRs.

## Validation


dgahagan:
Thanks — there is a good feature in here, and the care shows: grouping before
LIMIT/OFFSET so a long run takes one slot, refusing to guess when an item maps to two
Komga series, keeping `items_live` and the wishlist overlay intact. I am holding it, for
reasons that are mostly about where it sits rather than how it is written.

**It answers a general question for one provider.** "Collapse a series to one card in
Browse, and give it a page" is something every series in Shelf wants, not only Komga's. In
Shelf a series has no id — its *name* is its identity (`series_meta.name` is the primary
key, and every `/api/series/{name:path}/…` route is keyed by it). A per-series page on
that basis is already being designed. This PR adds a second, id-keyed series page at
`/series/komga/{id}` beside it, with its own copy of `series.find_gaps`. Two series models
is the thing I want to avoid.

**#133 already fixes the underlying complaint.** One manga run splitting into N series was
the real pain, and normalising the name fixes it inside the existing model — those volumes
now group correctly on `/series`.

Concrete problems, so the feedback is useful for the general version:

- **Grouped entries cannot be bulk-selected.** Select All collects `[data-item-id]`; the
  series card and row carry `data-komga-series-id` instead, and the row's select cell is
  empty. Komga items drop out of bulk move, tag and wishlist unless a series filter is on.
- **The heading count and the grid disagree.** "Collection (N)" still shows the raw item
  count while paging runs on the grouped count, so N is much larger than the cards shown.
- **It reroutes Browse for everyone.** Both `pages.browse` and `items.search_items` now go
  through `komga_series_browse.fetch_page`, including installs with no Komga. The added
  `, i.id ASC` tiebreak is a good idea, but it is a behaviour change for all users and
  deserves its own line in a changelog.
- The description says the representative cover prefers "the earliest numbered member with
  artwork"; `_series_meta` takes `rows[0].cover_path`, so an artless volume 1 gives a
  blank card.
- `_table_exists` is dead code — `komga_records` is created by the migrations on every
  install. And `app/services/komga_series_browse.py` imports from `app.routers.checkouts`,
  which points the dependency the wrong way; `OVERDUE_CONDITION` would need to move down.

I will post on this PR when the per-series page design settles, and I would value your
input on it then — particularly the grouping-before-pagination query, which is the hard
part and which you have already worked out.

dgahagan:
Deferring past **0.44.0**. This is the largest of your open PRs (725+/38− across 13 files) and it is also the one the generated-output problem has hurt most, so the order matters.

**What 0.44.0 changes.** Pull requests no longer carry generated build output — `static/css/app.css`, the `SW_VERSION` value in `static/sw.js`, and README's two test-count badges are regenerated on `main` after a merge. A new `generated-output` check refuses a PR that includes them and prints the exact restore command. Run `make css` and `make badges` locally to see your work; just leave the output out of the commit.

The collision this removes is the one you have been eating: both files are single lines, so any two PRs that touched a template conflicted on them regardless of content. With four of your PRs open at once, that has been most of your rebase cost.

**This PR** carries `static/css/app.css` and `static/sw.js`, and is `CONFLICTING`. After rebasing onto 0.44.0:

```
```

`app/routers/items.py` and the three shared card/row/grid fragments are the parts that may still need real reconciliation — the new Komga templates and services are new files and should come across untouched.

Grouping Komga series in Browse is a change to the main browse surface with a new drill-down page, so I want to give it a proper read against `browse_filters.py` (the filter set is declared once there and the `hx-include` lists, the SQL and `static/js/browse.js` all derive from it) before merging. That review is on me; nothing about the approach is in question.

dgahagan:
Deferred out of 0.45.0, and this one needs a rebase before it can go anywhere —
which is my fault, not yours.

`main` moved underneath it. 0.45.0 carries #131, which added a Source filter to
Browse, and a follow-up of mine (`4c4100c`) that moved that filter into the
Browse filter bar and reworked the surrounding item-grid fragments. That collides
directly with your `item_grid.html`, `item_cards_page.html` and
`item_rows_page.html` changes, and GitHub now reports the PR CONFLICTING. While
it is in that state GitHub cannot build a merge commit for it, so the checks you
see are from before the collision, not against current `main`.

One other thing to fix on the rebase: the diff carries `static/css/app.css` and
`static/sw.js`. Since 0.44.0 the `generated-output` job refuses generated output
on a pull request — those files collide across any batch of PRs, so they are
regenerated on `main` instead (and, as of 0.45.0, by CI rather than by hand).
Drop both from the branch and the job will pass; you do not need to run
`make css` at all.

Worth saying plainly: the grouping idea is one I want. The reshaping here is
release churn, not a verdict on the work. Rebase onto the 0.45.0 tag when it is
up and I will take it in the next cycle.

dgahagan:

Yesterday's comment still applies. This PR conflicts with `main` since 0.45.0 and needs a rebase before GitHub can run CI on it. 0.45.1 also touches `app/routers/items.py` and `app/routers/pages.py`, which this PR edits too, so rebase after the tag, not before, and you only have to do it once.

dgahagan:
Still deferred for 0.46.0, which ships Trash.

The earlier comment still applies: this PR conflicts with `main` and needs a rebase before GitHub can build it. One more thing to know when you rebase. After 0.46.0, every read of `items` and `item_copies` has to go through the `items_live` / `copies_live` views, so trashed items stay out of Browse and the series drill-down. `make check-deleted` flags a read of the physical tables. Once it's rebased and green, I'll take it in the next contributor batch.

dgahagan:
I am closing this, and I owe you a straight explanation, because my last few comments were misleading.

Since 0.44.0 I have been deferring this and telling you it needed a rebase, most recently "once it's rebased and green, I'll take it in the next contributor batch." That was wrong. The rebase was never the real blocker. The real blocker is the one in my first comment: this PR adds a second series model beside the one Shelf already has. It is keyed by the Komga series id and has its own page at `/series/komga/{id}`, while every other series in Shelf is keyed by its name. That objection still stands, and a rebase would not have changed it. I should have said so each time rather than let you think a rebase would get it merged. I'm sorry for the churn that caused.

What happens next: the general version — one card per series in Browse, and a page for each series showing what you own and what is missing — is still at the design stage on my side. It will be built on the name-keyed model, so it works for every series, not only Komga's. I will link that design here when it is ready for input, and I would value your review of it. Your grouping-before-pagination query solves the hard part, and I expect to borrow its shape. The concrete notes in my first comment (bulk-select on grouped cards, the heading count, the representative cover) are the checklist that version has to pass.

Thank you for the work here. Closing it is a decision about where the feature belongs, not about the quality of your code.

PR #133, opened 2026-09-19
Normalise Komga volume-suffixed series names

sudo-rpaisley:
## Summary

Komga can return series titles with an appended volume number, for example `One Piece (21)`. Shelf stores `seriesTitle` as `series_name`, so those values currently split one manga run into separate series groups.

This change:

- normalises numeric volume suffixes such as `One Piece (3)`, `One Piece (21)` and `One Piece (108)` to `One Piece`
- preserves four-digit publication-year run names such as `Batman (2016)`
- repairs previously stored volume-suffixed series names when the Komga record is resynchronised
- does not change the Komga schema, library selection or provider persistence model

## Tests

Passed the focused Komga suite, the full unit/static suite and the full Chromium E2E suite.


dgahagan:
Thanks — this is the right fix in the right place. Normalising in `normalise_book` means a
resync repairs rows Komga owns, and because non-Komga items only take `series_name` through
`_fill_missing_fields` when theirs is blank, nobody's hand-edited series name gets touched.
The year guard for `Batman (2016)` is the case I would have asked about, and it is tested.

One thing to drop before merge: the `README.md` hunk. The `Merge branch 'main'` commit took
your side of the badge lines, so the PR moves the counts *backwards* (3593 → 2986 unit,
256 → 230 E2E). CI is green only because the badge check is advisory on pull-request
builds; it fails on push to `main`. Reverting that one file is enough — badges get restamped
at release, so a PR never needs to touch them.

A known limit, fine to accept: a series genuinely titled `Something (2)` would lose its
suffix. I have not seen one in practice.

This also matters for #132 — see my note there.

dgahagan:
Deferring this one by a single release, and it is the closest of the four to going straight in — `MERGEABLE`, all three checks green, and the change itself (`app/services/komga_books.py` plus its two test files) is exactly the shape I want.

The one thing holding it is the README hunk. It bumps the two test-count badges from 3593/256 to your local numbers, and **0.44.0 makes that a hard refusal** on a `pull_request` build. Those badges, `static/css/app.css` and the `SW_VERSION` value in `static/sw.js` are now regenerated on `main` after a merge, so a PR should not carry any of them — a new `generated-output` check refuses one that does and prints the command to back it out. `make badges` is still right to run locally; just drop the result before you push.

So after 0.44.0 lands:

```
git checkout origin/main -- README.md
```

and this should merge on its own. If the badge bump is the *only* thing you drop, the diff goes to 3 files and nothing else about the PR needs to change.

Deferring rather than merging now because the release squash rewrites the README badge line anyway, so merging first would just hand you the conflict back. Thanks for keeping the volume-suffix normalisation self-contained — that made it easy to read.

dgahagan:
Merged as `d7aeac7a` — thank you. Dropping the README hunk was the only thing it needed, and you had it done within hours of 0.44.0. It ships in the next release; a Komga resync after upgrading will repair the rows that were split one-series-per-volume, and the `Batman (2016)` year guard means the runs that *should* stay apart do.

dgahagan:
Shipped in [v0.45.0](https://github.com/dgahagan/shelf/releases/tag/v0.45.0),
as you wrote it.

The four-digit-year exception is the part that makes this mergeable rather than
risky, and it was in your first version. Stripping a trailing `(N)` is easy;
knowing that `Batman (2016)` names a distinct run and must survive is the bit
that needs someone who has actually looked at a comics library. Without it this
would have silently merged two runs and there would have been no way back.

Also right: repairing already-split rows on the next sync instead of writing a
migration. A hand-edited series name on a non-Komga item is never touched, so
the blast radius stays inside what Komga owns.

Now documented under **Series names** in `docs/komga.md`, with the year
exception spelled out so nobody later "fixes" it.

Thanks — and the one before this, and the one after it.

PR #136, opened 2026-09-20
Add assisted periodical identification

sudo-rpaisley:
## Summary

Adds an assisted identification path for periodicals when a 977/ISSN scan does not return enough reliable metadata.

- keeps the normal 977/ISSN lookup as the first path;
- adds optional Google Books magazine-title assistance;
- selecting a candidate only refills the confirmation card and never adds an issue automatically;
- preserves the exact scanned 977 carrier and 2/5-digit supplement;
- accepts only checksum-valid confirmed ISSNs;
- keeps issue identity separate from publication identity;
- follows Shelf 0.44's current items_live and wishlist semantics;
- carries no generated CSS, service-worker stamp or README badge output.

## 0.44.0 port

This is one clean commit directly on current upstream `main`.

## Validation

The direct 0.44.0 validation passes the full Shelf unit/integration, static/offline and Playwright E2E workflow.


dgahagan:
Thanks — the design here is right, and it is the part that is easy to get wrong. The scanned 977 carrier and its add-on stay the issue identity, a selected candidate only refills the card, `normalise_issn` is checksum-strict, the volume id is regex-gated before it reaches a URL, and the lookups go through `outbound.fetch` on a host that is already paced. It also arrived with no generated output, a few hours after 0.44.0 asked for that. Three things before it merges, all small:

**1. The eBay link.** The card gains "Search exact barcode on eBay", always on. The public roadmap commits buy links to *off by default, per install* (`docs/roadmap.md`, under wishlist and series-gap rows), and I would rather not ship the first one as an exception to that. Please drop it. The plain web-search link I am fine with — it is a lookup, not a marketplace.

**2. The 4xx responses are invisible.** `/assist/search` returns 400 and `/assist/select` returns 404 as a bare `HTMLResponse`. HTMX does not swap a 4xx, so clicking **Use result** on a volume Google has since dropped does nothing at all — no message, no change. Return 200 with the message rendered into the target (the way the empty-results branch of `periodical_assisted_results.html` already does), or send `HX-Retarget`/`HX-Reswap` if you prefer to keep the status. #130 had the same shape and I should have said so more loudly there.

**3. `cover_url` is fetched and then dropped.** `assisted_periodical_select` puts it in `candidate`, but `periodical_scan_result.html` has no field for it, so `/api/periodicals/confirm` never sees it. Either carry it through as a hidden input and hand it to the cover queue on confirm, or stop collecting it. I would like the first — a magazine with a cover is the better outcome — but the second is a fine answer for this PR if the cover path is more than you want to take on here.

And one line in `docs/user-guide/periodicals.md` saying the assisted search exists and that it never adds anything by itself.

Two things I found while reading that are **not yours to fix**, recorded so you know I saw them:

- A corrected ISSN is not remembered against the carrier, so the next issue of the same magazine derives the same misleading 977 ISSN and the user repeats the assist. `upsert_publication` then matches on the corrected ISSN, so nothing duplicates. That is a reasonable v1.
- `periodical_records.upsert_publication` falls back to a title match and then overwrites that row's ISSN, so two editions that share a title (Wired US / Wired UK) merge. That predates this PR; yours only makes it easier to reach. I am filing it separately.

dgahagan:
Deferred out of 0.45.0 — the release closed its queue before this could get a
read. The gate is green, so this is about scope and timing, not the code.

What it waits on: the PR adds a router (`app/routers/periodicals.py`) and a new
outbound adapter (`app/services/periodical_google.py`) in one change. Both of
those are design-level surfaces here — a new route has to be walked past the
middleware order in `app/main.py` (`SecurityHeadersMiddleware` →
`RateLimitMiddleware` → `AuthMiddleware` → `CSRFMiddleware`), and a new metadata
source is expected to pace through `app/services/outbound.py` rather than calling
`httpx` directly. Neither is a criticism of what you wrote; they are just reads I
owe the change before it ships, and release eve is the wrong time to do them
quickly.

Timing: next cycle, reviewed properly. If the periodical work has a dependency
order you would rather I take it in, tell me and I will follow it.

dgahagan:

Heads-up: 0.45.1 changes `confirm_periodical_issue` in `app/routers/periodicals.py`, the router this PR also edits. When an issue is already known, the route now restores it from Trash before redirecting to it. After the tag, please merge `main` and check that your assisted path goes through the same `existing_id` branch and does not insert past it.

dgahagan:
Deferred once more, out of 0.46.0. I'm sorry this one keeps slipping.

0.46.0 ships Trash. Deleting an item or a copy now moves it to Trash instead of removing the row, and every read of `items` and `item_copies` goes through the `items_live` / `copies_live` views. `make check-deleted` enforces this. It's a large change to the delete and read paths, so I've kept this release to Trash plus the small Discogs follow-up (#141).

What this waits on: a rebase onto 0.46.0 once it's tagged. Check any new reads in `app/routers/periodicals.py` and `app/services/periodical_google.py` against `make check-deleted`. Then this is first in line for the next contributor batch, which I'm aiming for within the next couple of days.

dgahagan:
Merged and released in [0.47.0](https://github.com/dgahagan/shelf/releases/tag/v0.47.0). Thank you. This one waited through four releases for reasons that had nothing to do with your code, and I'm sorry it took this long.

All four review points landed as asked. The eBay link is gone, and your test now keeps any marketplace link off this card. Errors come back as 200 with `HX-Retarget`/`HX-Reswap`, so a stale volume shows a message instead of doing nothing. The docs line is in. You also rebased onto 0.46.0 the same morning it shipped, and you kept the Trash-restore path on `existing_id`, with a test for it. That saved me a round trip.

One difference from what I suggested: you took the "stop collecting it" answer on `cover_url`, so the thumbnail shows in the results but is not saved with the issue. That is fine for this PR, and the changelog says so, so nobody expects otherwise. Carrying it through to the cover queue is still a welcome follow-up if you want it.

Outside the PR itself, I added a paragraph to `docs/architecture.md` on the new Google Books adapter, and a line to the README's feature list and metadata-sources table.

PR #137, opened 2026-09-20
Add Discogs exact-release provider foundation

sudo-rpaisley:
## Summary

Adds the small provider foundation for optional Discogs exact-pressing enrichment of music records.

- searches Discogs `type=release` results rather than masters so candidates represent concrete pressings;
- supports title, artist, barcode and catalogue-number constraints;
- normalises Discogs Release/Master IDs, label/catalogue number, format details, genres/styles and collector identifiers such as matrix/runout and pressing-plant IDs;
- deduplicates equivalent identifiers;
- uses Shelf's existing outbound request/result abstractions;
- keeps MusicBrainz as Shelf's canonical release identity;
- does not alter Shelf item identity, schema, covers or item records;
- carries no generated build output under the 0.44 contribution rule.

## 0.44.0 port

This is one clean two-file provider commit directly on current upstream `main`. Credential configuration and exact-pressing persistence/UI are intentionally separate follow-ups.

## Validation

The direct 0.44.0 validation passes the full Shelf unit/integration, static/offline and Playwright E2E workflow.


dgahagan:
Thanks — this is a clean provider. It uses `outbound.fetch` and `provider_result` the way the other clients do, the token is a parameter rather than something the module goes looking for, searching `type=release` rather than masters is the right call for pressings, and the `identifiers` output already matches the existing `music_identifiers(identifier_type, value)` table — so the persistence follow-up has a home without a new table. Keeping MusicBrainz as the canonical identity is also exactly what I want.

One line before it merges:

**`api.discogs.com` needs a pacing entry.** `outbound.fetch` paces by host from `HOST_RATE_LIMITS` in `app/config.py`, and a host that is *absent* from that dict means **no pacing** — that default exists for LAN targets like RomM and Komga. Discogs allows 60 authenticated requests a minute, so:

```python
    # Discogs allows 60 authenticated requests per minute.
    "api.discogs.com": 1.0,
```

Nothing calls the module yet, so nothing is unpaced today — but the entry belongs with the client, not with whichever follow-up first reaches it.

Optional: the other clients send `Shelf/1.0 (+https://github.com/dgahagan/shelf)`; matching that format would be tidy.

One thing to look at before the UI follow-up rather than now: the Discogs API terms on attribution and on which fields an application may store. I have not read them closely enough to say there is a problem — only that the persistence PR is the point where it matters.

dgahagan:
Deferred out of 0.45.0 — the release is cutting now and this is landing after the
queue closed. Not a quality call: the gate is green and the two-file shape on
current `main` is exactly what makes a PR easy to take.

What it waits on is your own stated plan. You describe this as the foundation,
with credential configuration and exact-pressing persistence/UI as deliberate
follow-ups — so merging it alone ships `app/services/discogs.py` into a release
with nothing that calls it. I would rather land the provider and the thing that
uses it in the same version, so the changelog entry describes a capability a user
can reach rather than a file that exists.

Two things I noticed and want to keep: keeping MusicBrainz as the canonical
release identity and routing through the existing outbound abstractions are both
the right decisions, and they are the ones I would have asked for.

Rough timing: the release after 0.45.0, alongside whichever follow-up lands
first. If the follow-ups are going to be a while, say so and I will take this on
its own instead.

dgahagan:

Yesterday's reason still applies. This is the foundation for the Discogs work you described, so I would like to take it together with the first change that uses it, not by itself.

dgahagan:
Merged — thank you. It goes out in the next release.

One addition on my side, as flagged in the review: `api.discogs.com` now has a 1 req/s entry in `HOST_RATE_LIMITS` (`app/config.py`). A host that is not in that table gets no pacing at all, and Discogs allows 60 authenticated requests a minute. I added it at absorb time so this did not have to wait a third round.

The service still has no caller. The credential setting, persistence into `music_identifiers` and the UI are the follow-up. Before any UI ships, someone needs to read the Discogs API terms for attribution and for which fields we may store.

dgahagan:
Shipped in [v0.46.0](https://github.com/dgahagan/shelf/releases/tag/v0.46.0), together with #141's token setting. It's under Internal in the changelog, because nothing calls the provider yet.

PR #139, opened 2026-09-20
Add Dutch KB national bibliography provider

MartinZwolle:
## What this changes

Adds the Koninklijke Bibliotheek (KB) / Nederlandse Bibliografie Totaal (NBT)
as a national metadata provider for Dutch ISBN registration groups `978-90`
and `978-94`.

The provider queries the KB linked-data SPARQL endpoint and uses the existing
`bib_normalize` helpers for returned metadata.

## Why

Older Dutch books are often not found by Shelf's existing metadata sources.
NBT also has older records that contain only ISBN-10, while Shelf passes
ISBN-13 to national providers.

The KB provider therefore uses `canonical_isbn_pair()` and queries both forms
in a single SPARQL request:

```sparql
VALUES ?isbn { "<isbn13>" "<isbn10>" }

For example, 9789026101632 (Het transgalactisch liftershandboek by
Douglas Adams) is present in NBT as ISBN-10 9026101635, but not as its
ISBN-13.

## How it was tested

Recorded SPARQL JSON fixtures are used, so the tests do not require live
network access.

Test ISBNs include:

9789026101632 / 9026101635 — Het transgalactisch liftershandboek,
Douglas Adams, Fontein, 1980. NBT contains only the ISBN-10.
9789060695685 / 9060695682 — Wonderkinderen, Thea Beckman,
Lemniscaat. NBT returns multiple bibliographic records for this ISBN
(1984, 1989 and 1999), so the test verifies that a usable record is
returned without depending on result order.

The provider also has tests for empty results, malformed responses, HTTP
errors, rate limiting, transport failure and outbound pacing. Routing tests
cover both Dutch registration groups.

I also built the branch as a local Docker image and tested it in Shelf against
the live KB endpoint. Both ISBNs above were successfully populated.

 make test passes — 3656 passed
 make test-e2e passes
 make checks passes
 No generated output in this PR — static/css/app.css, static/sw.js's
SW_VERSION, and README's test-count badges are regenerated on main

I also ran make checks-fast, which passes.

## Notes for review

The KB/NBT SPARQL query retrieves title, linked author label, publication year,
language, page count, publication/publisher and ISBN.

data.bibliotheken.nl is added to outbound host pacing. No database or schema
changes are required.

The second fixture deliberately contains multiple NBT records for the same
ISBN. The provider follows the existing national-provider pattern and returns
the first usable record rather than trying to infer which physical edition the
user has.

---

  drop input.


dgahagan:
Thanks for this — and welcome. A Dutch national provider is a real gap, and the
ISBN-10/ISBN-13 detail you found is the kind of thing that only shows up against
a real catalogue: `canonical_isbn_pair()` with both forms in one `VALUES` clause
is the right shape, and recording SPARQL fixtures rather than calling out from
the tests is exactly the house convention.

**This is deferred out of 0.45.0, not rejected.** Two reasons, neither about the
code:

1. It arrived a few hours before the cut and no CI run has happened yet — a
   first-time contributor's workflow needs maintainer approval before it runs,
   so nothing has been verified by the public gate. Releasing an unbuilt provider
   is not something I'll do on release eve.
2. A new entry in `app/services/national.py` adds an outbound metadata source to
   the request path, which this project treats as a design-level change rather
   than a drop-in (pacing lives in `app/services/outbound.py`, with the per-host
   interval declared in `app/config.py`). I want to read it against that before
   it ships, not after.

You did the part contributors usually miss — `docs/architecture.md` and
`docs/configuration.md` are both in the diff. That is appreciated and it will
save the review time.

Next step on my side: approve the workflow run so the gate actually reports, then
review for the release after 0.45.0. I'll comment here either way rather than
leaving it silent.

MartinZwolle:
Thanks! Makes perfect sense, especially this close to the release. Happy to address anything that comes up in CI or review.

dgahagan:
Still deferred for 0.45.1, which is going out today. That release is internal groundwork for a future Trash feature: no new provider and nothing a user can see. I kept it to that one change on purpose.

Thanks for merging `main` into the branch. Your CI run is still waiting on my workflow approval, and that approval comes first. After CI is green, this PR is next in line for a proper review. It is not waiting on anything from you.

dgahagan:
Merged — thank you, Martin, and thanks for your patience through two deferrals. Dutch ISBNs (978-90 and 978-94) now go to the KB's Nederlandse Bibliografie Totaal first. It goes out in the next release.

I checked the SPARQL against the live endpoint before merging. `dct:` as `purl.org/dc/terms#` looked like a typo, but KB really does publish `issued` under that namespace, so your query is right. Your ISBN-10 fallback is a good catch too: the Douglas Adams record carries only `9026101635`.

One thing for later, not a blocker. 978-90 is shared with Flemish Belgium, so some of those ISBNs will not be in the NBT. That is harmless, because a `no_match` falls through to Open Library as usual.

MartinZwolle:
Thanks for the thorough review and for checking it against the live endpoint! Really pleased to see this merged. This was my first contribution to someone else's open-source project, so it was a fun one to learn from. And good point about Flemish 978-90 ISBNs — nice that the existing fallback handles those naturally.

dgahagan:
Shipped in [v0.46.0](https://github.com/dgahagan/shelf/releases/tag/v0.46.0). Dutch ISBNs now go to the KB first for everyone who pulls `latest`. It's in the changelog under Added, credited to you.

